PR #13, opened 2019-05-13
ask for rating in case CSAT is enabled

akshatdhabi:
- add asking for rating when the user has enabled CSAT ratings
- the rating will be taken in two parts first by smilies then user will have optional option to add a comment or restart a conversation

### How was the code tested?

- the code was tested by giving ratings to multiple chat threads


webplugin/css/app/mck-sidebox-1.0.css
```diff
@@ -5801,6 +5802,136 @@ div[name="message"]:last-child {
...
+    border: solid 0.5px #cecbcb;
+    max-width: 90%;
+    margin: 18px auto;
+}
+.mck-feedback-text-wrapper textarea{
+    background: none;
+    border: none;
+    outline: none;
+}
+.mck-feedback-text-wrapper textarea::-webkit-input-placeholder {
```
CaptainT33mo: @akshatdhabi add vendor prefixes for placeholder.

webplugin/js/app/kommunicate-ui.js
```diff
@@ -404,14 +404,46 @@ setAvailabilityStatus : function (status){
...
+            type: 'GET',
+            url: Kommunicate.getBaseUrl() + FEEDBACK_UPDATE_URL + '/'+ CURRENT_GROUP_DATA.tabId ,
+            global: false,
+            contentType: 'application/json',
+            success: function (data) {
+                CURRENT_GROUP_DATA.currentGroupFeedback = data.data;
+                kommunicateCommons.modifyClassList( {class : ["mck-box-form"]}, "n-vis");
+                kommunicateCommons.modifyClassList( {class : ["mck-csat-text-1"]}, "","n-vis");
+                kommunicateCommons.modifyClassList( {id : ["mck-sidebox-ft"]}, "mck-closed-conv-banner");
+                if( data.data && data.data.rating && data.data.comments.length> 0){
```
CaptainT33mo: @akshatdhabi don't you think that the entire code inside the `success` of the API call should be wrapped inside the `if(data.data && data.data.rating)` check?

Also, I think there should be an `error: function() { ... }` to handle any errors thrown by the API or code.

webplugin/js/app/kommunicate-ui.js
```diff
@@ -404,14 +404,46 @@ setAvailabilityStatus : function (status){
 showClosedConversationBanner  : function(isConversationClosed){
     var messageText = MCK_LABELS["closed.conversation.message"];
     var conversationStatusDiv = document.getElementById("mck-conversation-status-box");
-    if (isConversationClosed){
+    var isCSATenabled = kommunicate._globals.collectFeedback;
+    var FEEDBACK_UPDATE_URL = "/feedback"
```
devashishmamgain: @akshatdhabi if it is getting used only once then this can be removed.

webplugin/js/app/kommunicate-ui.js
```diff
@@ -404,14 +404,46 @@ setAvailabilityStatus : function (status){
...
+                if( data.data && data.data.rating && data.data.comments.length> 0){
+                    kommunicateCommons.modifyClassList( {id : ["csat-3","mck-rated"]}, "", "n-vis");
+                    kommunicateCommons.modifyClassList( {id : ["csat-1","csat-2"]}, "n-vis");
+                    document.getElementById('csat-3').innerHTML = '\"' + data.data.comments[0] + '\"';
+                }else if( data.data && data.data.rating) {
+                    kommunicateCommons.modifyClassList( {id : ["csat-2","mck-rated"]}, "", "n-vis");
+                    kommunicateCommons.modifyClassList( {id : ["csat-1","csat-3"]}, "n-vis");
+                    document.getElementById('mck-rating-container').innerHTML = kommunicateCommons.getRatingSmilies(data.data.rating);
+                }else {
+                    kommunicateCommons.modifyClassList( {id : ["csat-1"]}, "", "n-vis");
```
devashishmamgain: @akshatdhabi something looks wrong here, what if there are 5 type of ratings in future or lets say rating scale from 0 to 9, then do we need to write code for all 9 entries csat-2, csat-3, cast-4.... csat-9?

webplugin/js/app/kommunicateCommons.js
```diff
@@ -127,4 +127,24 @@ function KommunicateCommons() {
...
+    _this.isConversationClosedByBot = function(){
+        var filtered = CURRENT_GROUP_DATA.groupMembers.filter(function(member){
+            return member.userId == CURRENT_GROUP_DATA.lastMessagingMember
+        });
+        return filtered[0] && filtered[0].role == 2
+
+    }  
+
+    _this.getRatingSmilies = function(rating){
+        switch (rating) {
```
devashishmamgain: @akshatdhabi replace it with a json so that switch is not required.

webplugin/js/app/kommunicateCommons.js
```diff
@@ -127,4 +127,24 @@ function KommunicateCommons() {
         temporalDivElement.innerHTML = html;
         return temporalDivElement.textContent || temporalDivElement.innerText || "";
     }  
+    _this.isConversationClosedByBot = function(){
+        var filtered = CURRENT_GROUP_DATA.groupMembers.filter(function(member){
+            return member.userId == CURRENT_GROUP_DATA.lastMessagingMember
+        });
+        return filtered[0] && filtered[0].role == 2
```
CaptainT33mo: @akshatdhabi Semicolon, please. 😉 
Same thing below.

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -1775,6 +1779,63 @@ var MCK_MAINTAIN_ACTIVE_CONVERSATION_STATE;
...
+
+                restartConversation.addEventListener('click',function(){
+                    KommunicateUI.showClosedConversationBanner(false);
+                })
+
+                sendFeedbackComment.addEventListener('click',function(){
+                    feedbackObject.comments = [document.getElementById("mck-feedback-comment").value];
+                    feedbackObject.rating = CURRENT_GROUP_DATA.currentGroupFeedback.rating;
+                    feedbackObject.groupId = CURRENT_GROUP_DATA.tabId;
+                    mckUtils.ajax({
```
CaptainT33mo: @akshatdhabi van this `ajax` call and the one on the line `1821` be moved to a common function since both the code is almost identical?

webplugin/js/app/kommunicate-ui.js
akshatdhabi: These are not entries for the ratings, csat-1,2,3 are the containers of step 1,2,3.

devashishmamgain: one easier way can be:
remove n-vis from all before starting if else and then inside if else conditions, only modify that respective container.

akshatdhabi: The checks are there as the review can also be in parts, say ifa user has only rated the conversation but not commented on it we are supposed to show the comment box instead of final rating screen

webplugin/js/app/kommunicate-ui.js
akshatdhabi: No it will not be said `if(data.data && data.data.rating)` check as to when the data comes as null we will trigger the first step of CSAT rating

webplugin/js/app/kommunicateCommons.js
akshatdhabi: @devashishmamgain  done

webplugin/js/app/kommunicate-ui.js
akshatdhabi: @devashishmamgain  done

webplugin/js/app/kommunicate-ui.js
akshatdhabi: @devashishmamgain  done

PR #23, opened 2019-05-16
CO-483 Messages getting hidden in iOS browser [MASTER]

akshatdhabi:
- To fix message flicker in IOS devices


webplugin/css/app/mck-sidebox-1.0.css
```diff
@@ -1366,7 +1366,7 @@
 	right: 0;
 	bottom: 0px;
 	left: 0;
-	overflow: auto;
```
CaptainT33mo: @akshatdhabi why change `overflow: auto` to `overflow: visible`?

PR #28, opened 2019-05-16
[WIP] CO-513 Cookies and SessionStorage Error Fix [WIP]

CaptainT33mo:
-> Fix `Uncaught DOMException: Failed to read the 'sessionStorage' property from 'Window': Access is denied for this document.` error when cookies are disabled.

### How was the code tested?
1. Disable Cookies in the browser and load sidebox script.
2. Sidebox will load with no error in console.
3. A conversation will be created automatically on click of chat launcher and messages will be sent and received smoothly.

-> Type of `undefined` will not work in case of `sessionStorage` and `localStorage` if Cookies are disabled.

-> If cookies are blocked in the browser then our sidebox widget will not load as it is dependent on both Session Storage and Local Storage of the browse. And if cookies are disabled then both Session Storage and Local Storage will be inaccessible by the window object. 

## This PR is marked as WIP because it is dependent on `applozic.chat.min.js` file and we are using `sessionStorage` checks in that file as well. So @ab14bhardwaj merge this PR after you release the new version of `applozic.chat.min.js` file with the necessary fixes for `sessionStorage` checks.

### DIFF
--- a/webplugin/js/app/km-utils.js
+++ b/webplugin/js/app/km-utils.js
@@ -209,21 +209,27 @@ KommunicateUtils = {
         return text;
     },
     getDataFromKmSession: function (key) {
-        var session = sessionStorage.getItem(KommunicateConstants.KOMMUNICATE_SESSION_KEY);
-        return session ? JSON.parse(session)[key] : "";
+        if(KommunicateUtils.checkIsSessionStorageAvailable()) {
+            var session = sessionStorage.getItem(KommunicateConstants.KOMMUNICATE_SESSION_KEY);
+            return session ? JSON.parse(session)[key] : "";  
+        }
     },
     storeDataIntoKmSession: function (key, data) {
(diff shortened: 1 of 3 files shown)


webplugin/js/app/km-utils.js
```diff
@@ -209,21 +209,27 @@ KommunicateUtils = {
         return text;
     },
     getDataFromKmSession: function (key) {
-        var session = sessionStorage.getItem(KommunicateConstants.KOMMUNICATE_SESSION_KEY);
-        return session ? JSON.parse(session)[key] : "";
+        if(KommunicateUtils.checkIsSessionStorageAvailable()) {
```
devashishmamgain: @CaptainT33mo why 'check', how about KommunicateUtils.isSessionStorageAvailable()

CaptainT33mo: @devashishmamgain Renamed.

PR #31, opened 2019-05-18
CO-272 New message indicator in document's title

CaptainT33mo:
-> Whenever the current browser tab in which the script is running, is not in focus and a new message arrives then the current browser tab's title will say **`New message from <Sender Name>`**.

### How was the code tested?
-> By setting `messageTriggerTimeout` in `appOptions` and switching to different tab after reloading the tab with the script running.
-> By sending the message from the dashboard when the current tab is not in focus and widget is closed.

-> https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API

-> Fixed an issue where the chats not auto-scrolling when a new message arrives or when the user sends a message. This issue is already fixed and merged in release branch by @akshatdhabi but was not merged in master.


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -2033,6 +2035,8 @@ var CURRENT_GROUP_DATA={};
                             mckInitializeChannel.checkConnected(true);
                         }
                         _this.stopIdleTimeCounter();
+                        clearInterval(FLASH_PAGE_TITLE);
+                        parent.document.title = CURRENT_PAGE_TITLE;
```
devashishmamgain: @CaptainT33mo will this work in case of add kommunicate inside iframe?

CaptainT33mo: @devashishmamgain with `parent.document.title` it will not work for those people who are using our widget inside an iframe. And if I use `window.top` then it'll give `Blocked a frame with origin  from accessing a cross-origin frame.` error.

devashishmamgain:
Any other solution for this? There was something which Abhinav did for
image viewer, not sure if similar solution can be used here.

PR #56, opened 2019-06-18
Show CSAT rating in real time when a conversation is closed

akshatdhabi:
- Show the CSAT rating or Closed conversation banner in real time when a conversation is closed from the dashboard


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -9160,6 +9160,9 @@ var CURRENT_GROUP_DATA={};
                     var resp = $applozic.parseJSON(obj.body);
                     typeof resp.message == "object" && $mck_msg_inner.data('last-message-received-time', resp.message.createdAtTime);
                     var messageType = resp.type;
+                    if (typeof resp.message == "object" && resp.message.message === 'Status changed to Resolved'){
```
devashishmamgain: @akshatdhabi is there any other way to check than using "Status changed to Resolved'?
what if we later add localization and then text is changed to some other language.

akshatdhabi: @devashishmamgain  changed the check to use message metadata instead of the message

PR #69, opened 2019-06-27
voice input support for chrome

Suraj-Kumar:
->

### How was the code tested?

->

->

->


webplugin/pluginOptimizer.js
```diff
@@ -107,7 +107,11 @@ const compressAndOptimize = () => {
             path.resolve(__dirname, 'js/app/events/applozic-event-listener.js'),
             path.resolve(__dirname, 'js/app/events/applozic-event-handler.js'),
             path.resolve(__dirname, 'js/app/km-post-initialization.js'),
-            path.resolve(__dirname, 'js/app/mck-ringtone-service.js')
+            path.resolve(__dirname, 'js/app/mck-ringtone-service.js'),
```
devashishmamgain: @Suraj-Kumar is it possible to add these files at run time only for the customers who enable this feature through javascript code.

Suraj-Kumar: if we do that than it will append these file for every request which will increase API response time. this way it will append the file when server starts and response time remains constant. 
just a thought.

webplugin/js/app/media/media-service.js
```diff
@@ -0,0 +1,46 @@
...
+    },
+    processMicClickedEvent: function () {
+        if (!('webkitSpeechRecognition' in window)) {
+            alert("browser do not support speech recogization");
+        } else {
+            var recognition = new webkitSpeechRecognition();
+            recognition.continuous = false; // The default value for continuous is false, meaning that when the user stops talking, speech recognition will end. 
+            recognition.interimResults = true; // The default value for interimResults is false, meaning that the only results returned by the recognizer are final and will not change. Set it to true so we get early, interim results that may change. 
+            finalTranscript = '';
+            recognition.lang = "en-us";
```
CaptainT33mo: @Suraj-Kumar @devashishmamgain Will this language option be in any way related to the localization setting that we have in our chat widget. Also, we will be creating an option to add a welcome message in the dashboard in some of the languages other than English. So, shouldn't this thing support at least those languages? Just a thought.

Suraj-Kumar: got your point.  will check this.

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -445,6 +446,7 @@ var CURRENT_GROUP_DATA={};
         var USE_BRANDING = typeof appOptions.useBranding  == 'boolean'? appOptions.useBranding : true;
         var POPUP_WIDGET = appOptions.popupWidget;
         w.MCK_OL_MAP = new Array();
+        var VOICE_INPUT_ENABLED = appOptions.enableVoiceInput;
```
devashishmamgain: @Suraj-Kumar in options name, we can keep it simple as voiceInput. enable or disable is what its value will speak.

Suraj-Kumar: wanted to keep it in sync with other parameters.  other parameters use property true/false. shall  I change it?

devashishmamgain: @Suraj-Kumar no, i mean keep the property same as voiceInput, same as how other properties are locShare, popUpWidget, attachment etc.
voiceInput: true/false

Suraj-Kumar: ok

Suraj-Kumar: done

PR #71, opened 2019-06-28
Fix China CDN issue and launcher scrolling out of view issue [MASTER]

CaptainT33mo:
-> Changed CDN URL of jQuery file from `ajax.googleapis.com` to `cdnjs.cloudflare.com` to add support for China users.
-> Fixed an issue where the chat launcher icon was scrolling out of view in Android devices with Chrome version 75 while scrolling. This is an issue with the latest version of Chrome on Android. I've filed a bug on `bugs.chromium.org` and they've given a temporary workaround while they are fixing this issue and releasing it in next version. 

### DIFF
--- a/webplugin/css/app/mck-sidebox-1.0.css
+++ b/webplugin/css/app/mck-sidebox-1.0.css
@@ -2332,7 +2332,7 @@ display: none;
 
 .mck-sidebox-launcher {
 	white-space: pre;
-	position: fixed;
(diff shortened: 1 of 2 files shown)


jimaek:
Just so you know cdnjs is not blocked in China but its not allowed either. jsDelivr on the other hand has a Chinese ICP license and servers in mainland https://www.jsdelivr.com/network#china

PR #83, opened 2019-07-15
C-SAT bug fix, support for sessionTimeout via appOptions and replaced Jquery with JS.

ab14bhardwaj:
-> Remove commented links inside HTML file.
-> Replace jquery code inside mck-app.js with JS.
-> Fixed bug where C-SAT rating was showing in the wrong thread.
-> Added sessionTimeout support in appOptions(preference will be given to appOptions if sessionTimeout is coming from application settings API).
-> Removed Twilio related reference from the mck-app.js file.

### How was the code tested?

-> **sessionTimeout**
 Bypassing sessionTimeout : <integer value in milliseconds> in install script.

-> **C-SAT rating**
 By first opening two threads 1 and 2, then opened thread 2 and resolving thread 1 from the dashboard.

-> **Jquery->JS**
 By creating new build and checking that everything is working fine in our Sidebox.

-> For the real-time update of C-SAT rating the check was added inside onMessage function to detect the resolve event but it was not checking that which group is opened currently in sidebox. So if any conversation was getting resolved the C-SAT banner was appearing in the conversation which is currently opened (Other than the resolved one).

### DIFF
--- a/webplugin/js/app/mck-app.js
+++ b/webplugin/js/app/mck-app.js
@@ -130,7 +130,7 @@ function ApplozicSidebox() {
                 if (this.readyState == 4 && this.status == 200) {
                     var body = document.getElementsByTagName('body')[0];
                     body.innerHTML = this.responseText;
-                    var scriptContent = addScriptInstideHtml();
+                    var scriptContent = addScriptInsideHtml();
                     body.appendChild(scriptContent);
                     mckInitPluginScript();
                 }
@@ -145,7 +145,7 @@ function ApplozicSidebox() {
             return false;
         }
     };
-    function addScriptInstideHtml() {
+    function addScriptInsideHtml() {
         var script =  
         "function showAfterLoad(){"+
             "var mckSidebox = document.getElementById(\"mck-sidebox\");"+
--- a/webplugin/js/app/mck-app.js
+++ b/webplugin/js/app/mck-app.js
@@ -143,6 +143,7 @@ function ApplozicSidebox() {
             return false;
         }
     };
+            "var isScriptV2 = !!parent.document.getElementById('kommunicate-widget-iframe');"+
     function mckLoadStyle(url) {
         var head = document.getElementsByTagName('head')[0];
         var style = document.createElement('link');
--- a/webplugin/js/app/mck-app.js
+++ b/webplugin/js/app/mck-app.js
@@ -130,6 +130,8 @@ function ApplozicSidebox() {
                 if (this.readyState == 4 && this.status == 200) {
                     var body = document.getElementsByTagName('body')[0];
                     body.innerHTML = this.responseText;
+                    var scriptContent = addScriptInstideHtml();
+                    body.appendChild(scriptContent);
                     mckInitPluginScript();
                 }
             };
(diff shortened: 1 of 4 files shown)


webplugin/js/app/mck-app.js
```diff
@@ -332,6 +330,8 @@ function ApplozicSidebox() {
         try {
             var options = applozic._globals;
             var widgetSettings = data.chatWidget || data.widgetTheme;
+            var sessionTimeout = options.sessionTimeout;
+            sessionTimeout == null && (sessionTimeout = widgetSettings && widgetSettings.sessionTimeout);
```
devashishmamgain: @ab14bhardwaj priority should be given to the one passed from javascript

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -4466,6 +4468,7 @@ var CURRENT_GROUP_DATA={};
                 clearTimeout(MCK_TRIGGER_MSG_NOTIFICATION_PARAM);
                 MCK_MAINTAIN_ACTIVE_CONVERSATION_STATE && params.isGroup && params.tabId && KommunicateUtils.setItemToLocalStorage("mckActiveConversationInfo",{groupId:params.tabId, "appId":appOptions.appId, "userId":userId});
                 var currTabId = $mck_msg_inner.data('mck-id');
+                ACTIVE_TAB_ID = params.tabId || currTabId;
```
devashishmamgain: @ab14bhardwaj any other way to figure out the active tab id? so that we don't have to maintain the state of it and hence avoid any possible errors.

webplugin/js/app/mck-app.js
ab14bhardwaj: @devashishmamgain Priority is given to appOption only.
**var sessionTimeout = options.sessionTimeout;** This line is setting the value passed from SessionTimeout.
If nothing is passed or the value is null or undefined then it will check that API is returning any response or not.
**sessionTimeout == null && (sessionTimeout = widgetSettings && widgetSettings.sessionTimeout);**

webplugin/js/app/mck-sidebox-1.0.js
ab14bhardwaj: If not maintain then we have to get it from the UI from some other way. I guess maintaining the state would have been better rather than doing *$mck_msg_inner.data('mck-id');* in every function which we're doing currently. But no problem I will check for an alternate solution.

webplugin/js/app/mck-app.js
devashishmamgain: got it :)

webplugin/js/app/mck-sidebox-1.0.js
devashishmamgain: problem with having one single truth at multiple places is that then both should be syncd, even if one of them is missed somewhere by someone then bugs will appear. One way to have it store on the variable as long with data is to have a single function which does both the things ie storing it to ACTIVE_TAB_ID and storing it in data, and then all the code should call that common function without the need to set the data('mck-id'). this way it will be ensured that both are in sync always.

webplugin/js/app/mck-app.js
```diff
@@ -142,6 +145,35 @@ function ApplozicSidebox() {
             return false;
         }
     };
+    function addScriptInstideHtml() {
```
devashishmamgain: @ab14bhardwaj pls check spelling for Inside

webplugin/js/app/mck-app.js
```diff
@@ -142,6 +145,35 @@ function ApplozicSidebox() {
             return false;
         }
     };
+    function addScriptInsideHtml() {
+        var script =  
+        "function showAfterLoad(){"+
```
devashishmamgain: @ab14bhardwaj this looks confusing and difficult to maintain, any other solution for this? see if writing it as a separate function and then using function.toString() works?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/toString

webplugin/js/app/mck-app.js
```diff
@@ -146,30 +146,33 @@ function ApplozicSidebox() {
...
+                    }, false);
+                } else {
+                    window.addEventListener('kmInitilized', function () {
+                        showAfterLoad();
+                    }, false);
+                }
+            };
+        };
 
+        var script = String(scriptData) + "detectBrowserAndMakeUiVisible();"
```
ab14bhardwaj: Can't create an immediately invoked function because it will get executed when assigned to a variable or passed in String constructor it will get executed. That's why added the function call separately.

PR #102, opened 2019-07-30
CO-581 Accessibility Support in Chat Widget

CaptainT33mo:
-> Added accessibility support in the chat widget. This will increase the support of our application with assistive technologies like screen readers.

### How was the code tested?

-> By running NVDA Screen Reader app on Windows 10 Pro. 
-> Tested on Chrome, Firefox, Edge, Edge (new), IE11

->Accessible Rich Internet Applications (WAI-ARIA)
https://developer.mozilla.org/en-US/docs/Learn/Accessibility/WAI-ARIA_basics
https://www.w3.org/TR/wai-aria-1.1/#role_definitions
https://dequeuniversity.com/library/
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

Adding @parthsh as a reviewer to suggest any better text for `aria-label` attribute if any.


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -1378,9 +1378,9 @@ var CURRENT_GROUP_DATA={};
                 `+CUSTOM_CHAT_LAUNCHER+` 
                 </div>`
                   if (isAnonymousChat) {
-                      return '<a href="#" target="_self">' + CHAT_CLOSE_BUTTON + (CUSTOM_CHAT_LAUNCHER ? customLauncherHtml : defaultHtml) + '</a>';
+                      return '<a href="#" target="_self" aria-label="Open Kommunicate Chat" tabindex="0" role="button">' + CHAT_CLOSE_BUTTON + (CUSTOM_CHAT_LAUNCHER ? customLauncherHtml : defaultHtml) + '</a>';
```
devashishmamgain: @CaptainT33mo can we remove "kommunicate" as for end users of the customer, it wouldn't be "Kommunicate chat"

webplugin/plugin.js
```diff
@@ -72,6 +72,7 @@ function createKommunicateIframe() {
   kommunicateIframe.setAttribute("style", "overflow:hidden;"); // to fix scrollbars appearing before the chat widget loads on slow connections
   kommunicateIframe.setAttribute("scrolling", "no"); // to fix scrollbars appearing before the chat widget loads on slow connections
   kommunicateIframe.setAttribute("id", "kommunicate-widget-iframe");
+  kommunicateIframe.setAttribute("title", "Kommunicate live chat");
```
devashishmamgain: @CaptainT33mo remove "kommunicate"

webplugin/js/app/mck-sidebox-1.0.js
CaptainT33mo: @devashishmamgain we can remove it but we need to add some title here otherwise the screenreaders will just read out the class names and HTML tag names.

webplugin/plugin.js
CaptainT33mo: @devashishmamgain Should I just keep it `Live chat`?

devashishmamgain: yes we can keep "Live chat"

webplugin/js/app/mck-sidebox-1.0.js
devashishmamgain: we can keep as Chat

PR #104, opened 2019-08-05
CO-581 adding alt tags in profile image of agent in sidebox for accessibility

CaptainT33mo:
-> Adding `alt` attribute in image `` tag so that when navigating screenreaders should read the name of the image from the alt tag and not the image URL itself.

### How was the code tested?

-> By running NVDA and navigating through the chat widget.

->

->

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -4016,6 +4016,7 @@ var CURRENT_GROUP_DATA={};
             };
             _this.updateConversationHeader = function (params) {
                 var imageUrl;
+                var profileImage = params.name ? params.name + " profile image" : "Profile image";
                 $mck_tab_title.html(params.name);
                 $mck_tab_title.attr('title', params.name);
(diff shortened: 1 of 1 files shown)


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -5369,18 +5373,20 @@ var CURRENT_GROUP_DATA={};
             };
 
             _this.getImageUrlForGroupType = function (contact, displayName) {
-                return contact.imageUrl? '<img src="' + contact.imageUrl + '"/>' :  _this.getContactImageByAlphabet(displayName);
+                var profileDisplayName = displayName ? displayName + ' profile image' : 'Profile image';
```
devashishmamgain: @CaptainT33mo will it work if the display name contains special characters or if it contains a single quote ' such as  D'Souza

CaptainT33mo: @devashishmamgain yes. Not sure how the different screen readers will read it out aloud in their accents but yes it will work.

PR #109, opened 2019-08-07
TD-172: actionable message- form

jithinbalu:
-> create a form in  actionable message
### How was the code tested?

->
image

->

->

### DIFF
--- a/webplugin/css/app/km-rich-message.css
+++ b/webplugin/css/app/km-rich-message.css
@@ -1188,3 +1188,46 @@
 .km-cta-multi-button-container .km-quick-replies.km-carousel-card-button {
     border-color: #e7e4e4
 }
+.mck-form-template-wrapper input {
+    width: 93%;
+    padding: 9px;
+    margin-bottom: 2px;
+    display: inline-block;
+    border: none;
+    background-color: #ffffff;
+    color: #000000;
+    font-size: 10px;
+    margin-bottom: 12px;
+}
+.mck-form-template-wrapper label {
+    margin-bottom: 2px;
+    margin-left: 2px;
+}
+.mck-form-template-wrapper input::placeholder {
+    color: #929090;
+    font-size: 12px;
+}
+.mck-form-template-wrapper input:focus{
+  border-color:  #000000;
+  outline: none;
+}
+.mck-form-template-wrapper {
+    background-color: #e5e5ed;
+    border-radius: 3px;
+    padding: 16px;
+    margin-top: 5px;
+}
+.mck-form-template-wrapper form {
+    margin: 0px;
+}
+.mck-form-label {
+  font-size: 12px;
+  letter-spacing: 0.36px;
+  color: #333333;
+}
+.mck-form-template-container button {
+    width: 23%;
+    border-radius: 19px;
+    padding: 5px;
+    margin-top: 12px;
+}
--- a/webplugin/css/app/km-rich-message.css
+++ b/webplugin/css/app/km-rich-message.css
@@ -1207,6 +1207,22 @@
     color: #929090;
     font-size: 12px;
 }
+.mck-form-template-wrapper input::-webkit-input-placeholder {
+    color: #929090;
+    font-size: 12px;
+}
+.mck-form-template-wrapper input::-moz-placeholder {
+    color: #929090;
+    font-size: 12px;
+}
+.mck-form-template-wrapper input::-ms-input-placeholder {
+    color: #929090;
+    font-size: 12px;
+}
+.mck-form-template-wrapper input::-moz-placeholder {
+    color: #929090;
+    font-size: 12px;
+}
 .mck-form-template-wrapper input:focus{
   border-color:  #000000;
   outline: none;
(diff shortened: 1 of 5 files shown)


webplugin/css/app/km-rich-message.css
```diff
@@ -1187,4 +1187,47 @@
...
+    background-color: #ffffff;
+    color: #000000;
+    font-size: 10px;
+    margin-bottom: 12px;
+}
+.mck-form-template-wrapper label {
+    margin-bottom: 2px;
+    margin-left: 2px;
+}
+.mck-form-template-wrapper input::placeholder {
```
CaptainT33mo: @jithinbalu please add all the prefixes fr placeholder for different browsers as well. See this link: https://css-tricks.com/almanac/selectors/p/placeholder/

webplugin/js/app/km-richtext-markup-1.0.js
```diff
@@ -279,6 +279,21 @@ getCardInfoTemplate: function() {
             </div>
             <div class="km-carousel-card-sub-title">{{subtitle}}</div>
             <div class="{{cardDescriptionClass}}"><div class="km-carousel-card-description">{{description}}</div></div>`
+},
+getFormTemplate: function() {
+    return `<div class="mck-msg-box-rich-text-container mck-form-template-container">
+                <div class="mck-form-template-wrapper">
+                    <form id="mck-actionable-form" class="km-btn-hidden-form" action="{{actionUrl}}" method="post" target="_blank">
```
CaptainT33mo: @jithinbalu why `target="_blank"` on form?

webplugin/js/app/km-rich-text-event-handler.js
```diff
@@ -278,10 +278,11 @@ Kommunicate.richMsgEventHandler = {
 
         var target = e.target || e.srcElement;
         var requestType = target.dataset.requesttype;
-        var replyText = target.title || target.innerHTML;
-        var buttonType = target.dataset.buttontype;
+        var buttonType = target.dataset.buttontype || target.type;
         var data = {};
-        var form =target.parentElement.getElementsByClassName('km-btn-hidden-form')[0];
+        var form =target.parentElement.getElementsByClassName('km-btn-hidden-form')[0]
+        var isActionableForm = (form.id == "mck-actionable-form");
```
devashishmamgain: @jithinbalu will it be only one form? what if there are multiple forms in the chat history?

webplugin/js/app/km-rich-text-event-handler.js
```diff
@@ -305,14 +306,10 @@ Kommunicate.richMsgEventHandler = {
...
         }
-        if(replyText){
-            var messagePxy = {
-                'message': replyText, //message to send 
-            };
-    
-            Kommunicate.sendMessage(messagePxy);
-        }
-
+        var messagePxy = {};
```
devashishmamgain: @jithinbalu is it going to be set only for dialogflow case or for all whoever is using form?

webplugin/js/app/km-richtext-markup-1.0.js
jithinbalu: facing redirection issue

webplugin/js/app/km-rich-text-event-handler.js
jithinbalu: @devashishmamgain  the form object is created from parentElement of the clicked button (line 283)

devashishmamgain: @jithinbalu so at any given point of time, it will be always 1 element with id = "mck-actionable-form"?

webplugin/js/app/km-rich-text-event-handler.js
jithinbalu: @devashishmamgain  KM_CHAT_CONTEXT  will be added in the metadata if they did not set formActionUrl in the dialogflow JSON and messagePxy.message will be set if replyText is present (for all users)

webplugin/js/app/km-rich-text-event-handler.js
jithinbalu: @devashishmamgain ok got the point there should not be multiple forms with the same id. i will improve it

jithinbalu: @devashishmamgain "id" is removed from the form and id depended code also improved

webplugin/js/app/km-rich-text-event-handler.js
devashishmamgain: @jithinbalu @Suraj-Kumar currently we are assuming that if it is not URL then it is dialogflow chat context, can we have a different identifier for dialogflow chat context? in future we might have to build one for amazon lex and other platforms

webplugin/css/app/km-rich-message.css
jithinbalu: changes are done

webplugin/js/app/km-rich-text-event-handler.js
jithinbalu: @devashishmamgain @Suraj-Kumar "postBackToBotPlatform" identifier is added for bot platform chat context and
requestType can have 3 values
empty - submit a form on the action url
json - make a xhr request on action url
postBackToBotPlatform - send data back to bot platform.

PR #121, opened 2019-08-13
link button fix [Master]

jithinbalu:
->

### How was the code tested?

->

->

->

### DIFF
--- a/webplugin/js/app/km-rich-text-event-handler.js
+++ b/webplugin/js/app/km-rich-text-event-handler.js
@@ -279,11 +279,11 @@ Kommunicate.richMsgEventHandler = {
         var target = e.target || e.srcElement;
         var requestType = target.dataset.requesttype;
         var buttonType = target.dataset.buttontype || target.type;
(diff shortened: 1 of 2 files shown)


webplugin/js/app/km-rich-text-event-handler.js
```diff
@@ -279,13 +279,13 @@ Kommunicate.richMsgEventHandler = {
         var target = e.target || e.srcElement;
         var requestType = target.dataset.requesttype;
         var buttonType = target.dataset.buttontype || target.type;
+        if(buttonType !="submit"){   
+            return ;
+        }
         var data = {};
         var form =target.parentElement.getElementsByClassName('km-btn-hidden-form')[0]
         var isActionableForm = (form.className.indexOf("mck-actionable-form") != -1 );
```
devashishmamgain: @jithinbalu safe side, can we add a check for form exists?

jithinbalu: @devashishmamgain  changes are done

PR #135, opened 2019-08-22
CO-773 Removed outline when clicking on any elements in sidebox chat widget

CaptainT33mo:
-> Removed outline when clicking on chat launcher, messages or any conversations. Instead, added event listeners only when someone is navigating the chat widget using arrow keys.

### How was the code tested?

-> By navigating the chat widget using arrow keys and mouse.

### DIFF
--- a/webplugin/css/app/mck-sidebox-1.0.css
+++ b/webplugin/css/app/mck-sidebox-1.0.css
@@ -6225,3 +6225,10 @@ div#mck-rated:before{
 	  opacity: 0;
 	}
   }
+
+body:not(.accesibility) :focus {
+	outline: none;
+}
+body.accesibility :focus {
+	outline: 5px auto #3b99fc;
+}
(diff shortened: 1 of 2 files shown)


webplugin/css/app/mck-sidebox-1.0.css
```diff
@@ -6224,4 +6224,11 @@ div#mck-rated:before{
 	  -webkit-transform: scale(2.4);
 	  opacity: 0;
 	}
-  }
\ No newline at end of file
+  }
+
+body:not(.accesibility) :focus {
```
devashishmamgain: Can this be restricted to only kommunicate specific element assuming that this is required for both v1 and v2

CaptainT33mo: @devashishmamgain are we still supporting V1? If someone faces this outline issue in V1 then we can tell him to use V2.

devashishmamgain:
Currently we have an old build deployed for v1, if it is easy to modify it
to work with both then we can do it, if it's complicated then we can ignore
v1

webplugin/css/app/mck-sidebox-1.0.css
CaptainT33mo: > Currently we have an old build deployed for v1, if it is easy to modify it
to work with both then we can do it, if it's complicated then we can ignore
v1
@devashishmamgain if we try to modify this specifically for V1 then we need to add javascript code for most of the selectors of our sidebox. If any V1 users face this issue related to outline and don't want to move to V2 then we can share 1 line CSS fix which he can directly add it in his website.

devashishmamgain:
Ok

PR #152, opened 2019-09-03
Update logout function and remove all script tags on logout.

ab14bhardwaj:
### How was the code tested?

> By calling Kommunicate.logout() and reinitializing Kommunicate from console.

### DIFF
--- a/webplugin/js/app/kommunicate.js
+++ b/webplugin/js/app/kommunicate.js
@@ -243,6 +243,7 @@ $applozic.extend(true,Kommunicate,{
             window.$applozic.fn.applozic('logout');
         };
         KommunicateUtils.deleteUserCookiesOnLogout();
+        parent.window && parent.window.removeKommunicateScripts();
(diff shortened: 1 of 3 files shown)


devashishmamgain:
@ab14bhardwaj whats the purpose of removing the scripts and then adding it back?
what if we check internally if the script objects are already present then call some login function internally without loading the scripts.

ab14bhardwaj:
> what's the purpose of removing and then adding the scripts again? need to find a better way to achieve it, one way can be to have an internal check if the script objects are already present then directly make the call to the login function (if login function is not there, we can create one) without the need to reload the js.

- A customer want's to pass some new appOptions on login. For example, locShare : true then how would we load google maps file?
Should we allow the customer to pass appOptions on re-login and if yes then I guess we need to delete the iframe or check every script tag before creating it ?

devashishmamgain:
good point 👍

PR #157, opened 2019-09-04
CO-734 Comment code which was replacing older cookies with new structure.

ab14bhardwaj:


Suraj-Kumar:
maybe we need it in the future?

ab14bhardwaj:
> maybe we need it in the future?

Then maybe we can comment out the code. Is it okay ?

PR #185, opened 2019-09-17
[Dev]fix: sentry is not catching the error

jithinbalu:
- Catch the error to sentry from chat widget.
- Change the sentry DSN for prod.
- Update the sentry SDK version 
### How was the code tested?

- created an error, the error is getting cathed into sentry.
an HTML file is created with an error and chat widget is added then deployed through Netlify. the error in the HTML file is not catching in the sentry. only widget errors caught.


server/config/config-env.js
```diff
@@ -85,7 +85,7 @@ const configEnv = {
     "thirdPartyIntegration": {
       "sentry": {
         "plugin": {
-          "dsn": "https://a7fe0d3754264f649630801e7349da27@sentry.io/1324243",
+          "dsn": "https://9f71614ef8184d0cab00074555dad9a7@sentry.io/1321911",
           "enable": true
         },
         "server": {
```
ab14bhardwaj: @jithinbalu Can we remove server DSN from the config file for all the environments as it is of no use here?

jithinbalu: @ab14bhardwaj server DSN removed from the config file

PR #187, opened 2019-09-17
CO-708 Improved image loading in launcher icon

CaptainT33mo:
-> Improved the load time of the custom launcher icon in the chat widget. Initially, it was taking time to load even if the image is smaller in size. 
Now the launcher icon will be visible after the image is loaded and also it'll appear with a casual scale animation to give a smoother experience.

### How was the code tested?

-> By loading the launcher with custom image set as its launcher icon.


webplugin/js/app/mck-app.js
```diff
@@ -416,6 +417,23 @@ function ApplozicSidebox() {
...
         }
     };
+
+
+    function preLoadLauncherIcon(widgetTheme) {
+        if(widgetTheme && widgetTheme.widgetImageLink) {
+            var img = new Image();
+            img.onload = function () {
+                var launcherInterval = setInterval(function(){
+                    if(document.getElementById("mck-sidebox-launcher")){
```
devashishmamgain: It must be present in some variable, we can reuse that, if not we can create one, this way it wouldn't be searching in DOM multiple times for same element

CaptainT33mo: @devashishmamgain  Actually it is needed to be searched in the DOM because when this code fires for the first time `mck-sidebox-launcher` is not present because it is appended in the DOM via javascript. But as soon as it is fired for the second time then at that time, it is present there.

I'll add it in a variable but that variable needs to be inside the `setInterval` so that it can check whether it is appended in DOM or not. So the calculate will happen again when the `setInterval` fires for the second time. And in my numerous testing it never fired more than once so there won't be any major performance issues.

PR #189, opened 2019-09-18
CO-804 Fixed image on Top Bar will be similar to the image visible in message sent by agent.

CaptainT33mo:
### How was the code tested?
 
-> Tested by starting a conversation after the agent's image is removed and also when a new image is set.

-> Issue is explained via the screenshots given below.

#### Before:
image

#### After:
image

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -4095,7 +4095,7 @@ var MCK_CHAT_POPUP_TEMPLATE_TIMER;
                 } else {
                     //DEFAULT_PROFILE_IMAGE
                     imageUrl = "";
-                    $applozic(".mck-agent-image-container .mck-agent-image-svg-container").html(kommunicateCommons.getDefaultAvatarImageSvg());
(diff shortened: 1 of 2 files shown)


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -5483,23 +5483,23 @@ var MCK_CHAT_POPUP_TEMPLATE_TIMER;
       			_this.getContactImageLink = function(contact, displayName, message) {
                         var imgsrctag = '';
                         var profileDisplayName = displayName ? displayName + ' profile image' : 'Profile image';
-                        if(!contact.isGroup){
+                        if(contact && !contact.isGroup){
```
devashishmamgain: @CaptainT33mo is contact && check required here? i guess if it is not coming, nothing will work in code below it, we can probably figure out why contact is coming as undefined and fix that part.

CaptainT33mo: @devashishmamgain actually I am sending `null` in that function like this: `_this.getContactImageLink(null, name)`
So, in that case, it'll throw error `Cannot read property isGroup of undefined`.

PR #197, opened 2019-09-23
fix classlist of undefined

akshatdhabi:


ChangXiaoning:
Hi, I am a Node.js learner. I am curious about the reason why document.getElementById('mck-char-warning') is undefined. Look forward for your reply. Thx.

PR #211, opened 2019-09-25
 Updating bots language code dynamically 

Suraj-Kumar:
->

### How was the code tested?

->

->

->

### DIFF
--- a/webplugin/js/app/km-richtext-markup-1.0.js
+++ b/webplugin/js/app/km-richtext-markup-1.0.js
@@ -143,12 +143,12 @@ getButtonTemplate:function(options,requestType, buttonClass){
 getQuickRepliesTemplate:function(){
     return`<div class="km-cta-multi-button-container">
             {{#payload}}
-                 <button aria-label="{{title}}" title='{{message}}' class="km-quick-replies km-custom-widget-text-color {{buttonClass}} " data-metadata = "{{replyMetadata}}">{{title}}</button>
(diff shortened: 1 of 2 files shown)


webplugin/js/app/km-richtext-markup-1.0.js
```diff
@@ -143,12 +143,12 @@ getButtonTemplate:function(options,requestType, buttonClass){
 getQuickRepliesTemplate:function(){
     return`<div class="km-cta-multi-button-container">
             {{#payload}}
-                 <button aria-label="{{title}}" title='{{message}}' class="km-quick-replies km-custom-widget-text-color {{buttonClass}} " data-metadata = "{{replyMetadata}}">{{title}}</button>
+                 <button aria-label="{{title}}" title='{{message}}' class="km-quick-replies km-custom-widget-text-color {{buttonClass}} " data-metadata = "{{replyMetadata}}" data-languageCode = "{{updateLanguage}}">{{title}}</button>
```
devashishmamgain: @Suraj-Kumar below it is {{action.updateLanguage}} here it is {{updateLanguage}}, both are supposed to be different or same?

Suraj-Kumar: both are different depends on JSON being passed.

devashishmamgain: is it json format specified by us somewhere or is it through dialogflow, trying to understand what makes this different than getGenericSuggestedReplyButton.
If it is enforced by us then why can't keep it consistent?

Suraj-Kumar: getGenericSuggestedReplyButton  was added to support different type of buttons in single message. So the JSON format is slightly different there. once this template is added into mobile SDKs will remove old JSON from docs and promote this one.

PR #227, opened 2019-10-03
new conversation fix and disable bug

hardikparashar:
->new conversation fix and disable bug

### How was the code tested?

->

->

->


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -4899,6 +4899,14 @@ var MCK_BOT_MESSAGE_QUEUE = [];
                         mckMessageService.getAndSetAwayMessage({}, params, conversationAssigneeDetails, isAgentOffline);
                         mckMessageService.updateConversationHeader(updateConversationHeaderParams);
                     });
+                    for (var key in params.groupDetails.users) {
```
devashishmamgain: @hardikparashar is it mandatory that params. groupDetails will always have some value?

PR #254, opened 2019-10-10
Remove unused code, combine all JS files to one and removed googleApiKey getting added twice.

ab14bhardwaj:
### How was the code tested?

> - By running the script and checking everything is working fine or not.


webplugin/pluginOptimizer.js
```diff
@@ -124,22 +124,24 @@ const compressAndOptimize = () => {
                 drop_console: true
             }
         },
-        output: path.resolve(__dirname, `${buildDir}/kommunicate-plugin.${version}.min.js`),
+        output: path.resolve(__dirname, `${buildDir}/kommunicate-plugin.min.js`),
```
devashishmamgain: @ab14bhardwaj is ${version} required here? if not, will it handle the browser cache issue?

ab14bhardwaj: @devashishmamgain It is not required here because the above file is not getting served anywhere.

PR #255, opened 2019-10-10
CO-841 Pre-chat popup will now work with pre-chat lead collection

CaptainT33mo:
-> Pre-chat popup widget will now work with Pre-chat lead collection.

### How was the code tested?

-> By enabling pre-chat popup and adding pre-chat lead collection code inside `onInit: function() { ... }` in kommunicate install script.

### DIFF
--- a/webplugin/js/app/km-message-markup-1.0.js
+++ b/webplugin/js/app/km-message-markup-1.0.js
@@ -24,18 +24,24 @@ Kommunicate.messageTemplate = {
 
 Kommunicate.popupChatTemplate = {
 
-    getPopupChatTemplate: function(popupWidgetContent, widgetTheme) {
+    getPopupChatTemplate: function(popupWidgetContent, widgetTheme, isAnonymousChat) {
         var isPopupEnabled = kommunicateCommons.isObject(widgetTheme) && widgetTheme.popup;
         var chatPopupTemplateMarkup = "";
         var popupMessageContent = popupWidgetContent && popupWidgetContent.length && popupWidgetContent[0].message;
         
         if(isPopupEnabled) {
+            var launcherClass = isAnonymousChat ? "km-anonymous-chat-launcher" : "applozic-launcher";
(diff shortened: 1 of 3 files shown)


webplugin/js/app/km-message-markup-1.0.js
```diff
@@ -24,18 +24,24 @@ Kommunicate.messageTemplate = {
...
             var templateCss = index === KommunicateConstants.CHAT_POPUP_TEMPLATE.HORIZONTAL ? 'chat-popup-widget-container--horizontal' : 'chat-popup-widget-container--vertical';
-            chatPopupTemplateMarkup = '<div id="chat-popup-widget-container" class="chat-popup-widget-container ' + templateCss + ' n-vis applozic-launcher"><div class="chat-popup-widget-text-wrapper"><p class="chat-popup-widget-text">' + (popupMessageContent && kommunicateCommons.formatHtmlTag(popupMessageContent)) + '</p></div></div>';
+            chatPopupTemplateMarkup = '<div id="chat-popup-widget-container" class="chat-popup-widget-container ' + templateCss + ' n-vis ' + launcherClass + '"><div class="chat-popup-widget-text-wrapper"><p class="chat-popup-widget-text">' + (popupMessageContent && kommunicateCommons.formatHtmlTag(popupMessageContent)) + '</p></div></div>';
         };
         
-
-        return chatPopupTemplateMarkup;
+        if(isAnonymousChat) {
+            var anonymousLauncherContainer = document.getElementById('km-anonymous-chat-launcher');
```
devashishmamgain: @CaptainT33mo any return required inside if?

CaptainT33mo: @devashishmamgain  No, no return statement is required here. It'll directly append the markup to DOM.

devashishmamgain: got confused as else is returning something, is that part required?

CaptainT33mo: @devashishmamgain The `else` part is returning string which contains HTML markup which will be used for the Pre-chat popup template which will be shown in normal cases without the Pre-chat lead collection. The code inside the `if` part will return DOM node which will directly be appended in our iframe in case of Pre-chat lead collection hence no return statement is needed there.

PR #268, opened 2019-10-14
CO-811 Hide iframe on login failure and minor code simplification.

ab14bhardwaj:
-> Hide iframe on login failure and minor code simplification.

### How was the code tested?

-> kommunicate-hide-custom-iframe class will be added on Iframe in case of login failure to hide it.

### DIFF
--- a/webplugin/js/app/kommunicate.js
+++ b/webplugin/js/app/kommunicate.js
@@ -501,10 +501,10 @@ $applozic.extend(true,Kommunicate,{
         };
     },
     /**
-     * @param {Boolean} hide 
+     * @param {Boolean} display 
      */
-    showKommunicateWidget: function(hide) {
+    displayKommunicateWidget: function(display) {
         var kommunicateIframe = parent.document.getElementById('kommunicate-widget-iframe');
-        hide ? kommunicateIframe.classList.remove("kommunicate-hide-custom-iframe") : kommunicateIframe.classList.add("kommunicate-hide-custom-iframe");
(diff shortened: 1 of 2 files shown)


webplugin/js/app/kommunicate.js
```diff
@@ -501,10 +501,10 @@ $applozic.extend(true,Kommunicate,{
...
     },
     /**
-     * @param {Boolean} hide 
+     * @param {Boolean} display 
      */
-    showKommunicateWidget: function(hide) {
+    displayKommunicateWidget: function(display) {
         var kommunicateIframe = parent.document.getElementById('kommunicate-widget-iframe');
-        hide ? kommunicateIframe.classList.remove("kommunicate-hide-custom-iframe") : kommunicateIframe.classList.add("kommunicate-hide-custom-iframe");
+        display ? kommunicateIframe.classList.remove("kommunicate-hide-custom-iframe") : kommunicateIframe.classList.add("kommunicate-hide-custom-iframe");
```
devashishmamgain: @ab14bhardwaj if someone calls displayKommunicateWidget(false) two times and then calls displayKommunicateWidget(true) will it work?

ab14bhardwaj: @devashishmamgain yes it will work because classList.add won't add same class twice even if you execute the line  "classList.add("kommunicate-hide-custom-iframe")". So even after n number of attempts you call "classList.add("kommunicate-hide-custom-iframe")" it will work and remove the class and show our widget.

PR #283, opened 2019-10-21
fix: {url} has been blocked by CORS policy while form submission

jithinbalu:
-> fix: {url} has been blocked by CORS policy while form submission. not for all url's , zaper webhook is throwing an error.

### How was the code tested?

->  from tested with action URL as zapier webhook, data is submitting to webhook 

->

-> currently we are using  Applozic.ALApiService.ajax() method to make POST request . after replacing this method with javascript XMLHttpRequest, data is submitting all URL's(zapier webhook) without any  CORS policy issue. 

### DIFF
--- a/webplugin/js/app/km-rich-text-event-handler.js
+++ b/webplugin/js/app/km-rich-text-event-handler.js
@@ -292,18 +292,16 @@ Kommunicate.richMsgEventHandler = {
             data[inputs[i].name] = inputs[i].value; 
         }
         if (requestType == "json") {  
-           KommunicateUtils.isURL(form.action) && window.Applozic.ALApiService.ajax({
-            url: form.action,
-            async: false,
-            type: "post",
-            data:JSON.stringify(data),
-            contentType:"application/json",
-            success: function (data) { 
-            },
-            error: function (xhr,desc, err) {
-               console.log("error while sending data ",err); 
+        var xhr = new XMLHttpRequest();
+        xhr.onreadystatechange = function() {
+            if (this.readyState == 4 && this.status == 200) {
+            // success response    
+            // console.log(this.responseText);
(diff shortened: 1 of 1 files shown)


webplugin/js/app/km-rich-text-event-handler.js
```diff
@@ -292,18 +292,16 @@ Kommunicate.richMsgEventHandler = {
...
-            contentType:"application/json",
-            success: function (data) { 
-            },
-            error: function (xhr,desc, err) {
-               console.log("error while sending data ",err); 
+        var xhr = new XMLHttpRequest();
+        xhr.onreadystatechange = function() {
+            if (this.readyState == 4 && this.status == 200) {
+            // success response    
+            // console.log(this.responseText);
```
devashishmamgain: remove console log

webplugin/js/app/km-rich-text-event-handler.js
```diff
@@ -292,18 +292,16 @@ Kommunicate.richMsgEventHandler = {
             data[inputs[i].name] = inputs[i].value; 
         }
         if (requestType == "json") {  
-           KommunicateUtils.isURL(form.action) && window.Applozic.ALApiService.ajax({
```
jithinbalu: @ab14bhardwaj 
Some API calls are not happening due to CORS restriction, this error is happening with Applozic.ALApiService.ajax() method. after replacing that method with javascript XMLHttpRequest() POST request is worked. So do we have any CROS restrictions at chat-widget?

devashishmamgain: in the common ajax method, we are adding our own request headers which our servers require. for external calls, we can avoid using the common code.

PR #294, opened 2019-11-08
fix CSAT ratings failing in caase of Dialogflow close conversation 

akshatdhabi:
- Fix the error that was thrown and CSAT banne not showing up incase the conversation is closed by diaglogflow payload

### DIFF
--- a/webplugin/js/app/kommunicateCommons.js
+++ b/webplugin/js/app/kommunicateCommons.js
@@ -131,11 +131,14 @@ function KommunicateCommons() {
         return html.replace(/</g, '&lt;').replace(/>/g, '&gt;');
     };
     _this.isConversationClosedByBot = function () {
-        var filtered = CURRENT_GROUP_DATA.groupMembers.filter(function (member) {
-            return member.userId == CURRENT_GROUP_DATA.lastMessagingMember;
-        });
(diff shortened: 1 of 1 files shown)


webplugin/js/app/kommunicateCommons.js
```diff
@@ -131,11 +131,14 @@ function KommunicateCommons() {
         return html.replace(/</g, '&lt;').replace(/>/g, '&gt;');
     };
     _this.isConversationClosedByBot = function () {
-        var filtered = CURRENT_GROUP_DATA.groupMembers.filter(function (member) {
-            return member.userId == CURRENT_GROUP_DATA.lastMessagingMember;
-        });
-        return filtered[0] && filtered[0].role == KommunicateConstants.APPLOZIC_USER_ROLE_TYPE.BOT;
-
+        if (CURRENT_GROUP_DATA.groupMembers && Array.isArray(CURRENT_GROUP_DATA.groupMembers)) {
```
devashishmamgain: @akshatdhabi this will avoid the error or will fix the functionality?

akshatdhabi: the function is called twice in this particular case, once when the message is being populated. When that happens the CURRENT_GROUP_DATA object is already poputad.

PR #303, opened 2019-11-16
CO-869 Added appOption to disable pre chat popup and also to set custom pre chat popup messages

CaptainT33mo:
-> From now on if someone wants to override any of the parameters coming from the appSettings API like pre-chat popup or bot message delay or any of the widget customization settings then he can pass those settings' value in the `appSettings` object inside the kommunicate install script as an appOption.

**For Example:**
If someone wants to disable the pre-chat popup on a page from install script then just pass the following in the KM install script as an appOption:
```
"appSettings": {
    chatWidget: {
        popup: false,
    }
}
```

### How was the code tested?

-> By passing the following object in the kommunicate install script
```
"appSettings": {
    chatWidget: {
        popup: false,
    },
    chatPopupMessage: [{
        message: "Hello"
    }]
}
```


CaptainT33mo:
> is it possible that we provide a complete app settings object in appOptions? then we will not need to develop or add any additional parameter for any field.
> final settings will be 'appSettings' json passed in appOptions + app settings coming from server.
> @CaptainT33mo @ab14bhardwaj

@devashishmamgain @ab14bhardwaj I've done the changes. From now on if someone wants to change the message of pre-chat popup then he will have to add the following in the kommunicate install script:

```
"appSettings": {
    chatWidget: {
        popup: false,
    },
    chatPopupMessage: [{
        message: "Hello"
    }]
},
```

CaptainT33mo:
> existing released app options on the docs will still work?

@devashishmamgain Yes, the existing appOptions mentioned in the docs will continue to work as it is. We are only changing those appOptions that are coming from the appSettings API.

PR #306, opened 2019-11-21
Upload build files to CDN and then use them.

ab14bhardwaj:
### How was the code tested?

> - By running the node server with below command and upload files to the CDN-kommunicate bucket for the test AWS account.
`npm run dev -- --upload`

- https://www.npmjs.com/package/node-fetch
- https://www.npmjs.com/package/minimist
- How to pass arguments even though you have declared what script will do in package.json?
> Solution : *npm run dev* command will execute *"node ./server/app.js"*. So I have to pass the arguments from the command we wrote initially to the root command then I need to write the script like below.
`npm run dev -- --upload` Here --upload is the argument.


webplugin/pluginOptimizer.js
```diff
@@ -225,6 +239,20 @@ const generateFilesByVersion = (location) => {
...
+    try {
+        fs.unlinkSync(path);
+    } catch (error) {
+        console.log(error);
+    }
+};
+
+const uploadFilesToCdn = async(buildDir, version) => {
+    await pluginClient.upload(buildDir, version);
+    console.log("Uploaded all file to CDN");
```
devashishmamgain: files

ab14bhardwaj: Done.

PR #332, opened 2019-12-18
fix: error while uploading image

jithinbalu:
-> fix: error while uploading the image.

### How was the code tested?

-> upload image -> stop upload -> immediately click on re-upload - > no errors are throwing

->

-> Error is throwing when the user tries to re-upload the image immediately after cancelling the upload. Image upload API is taking time(5sec) to return fileMeta Object.

### DIFF
--- a/webplugin/js/app/km-rich-text-event-handler.js
+++ b/webplugin/js/app/km-rich-text-event-handler.js
@@ -35,9 +35,10 @@ Kommunicate.attachmentEventHandler= {
         var uploadIconHidden= $applozic(e.target).closest('.km-msg-box-attachment').find('.km-progress-upload-icon').hasClass('n-vis');
         var attachmentDiv= $applozic(e.target).closest('.km-msg-box-attachment').children();
         var msgkey = attachmentDiv[0].dataset.msgkey;
-        var deliveryStatusDiv= $applozic(e.target).closest('.mck-clear').find('.mck-msg-right-muted');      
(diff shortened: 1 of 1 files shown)


webplugin/js/app/km-rich-text-event-handler.js
```diff
@@ -35,9 +35,10 @@ Kommunicate.attachmentEventHandler= {
         var uploadIconHidden= $applozic(e.target).closest('.km-msg-box-attachment').find('.km-progress-upload-icon').hasClass('n-vis');
         var attachmentDiv= $applozic(e.target).closest('.km-msg-box-attachment').children();
         var msgkey = attachmentDiv[0].dataset.msgkey;
-        var deliveryStatusDiv= $applozic(e.target).closest('.mck-clear').find('.mck-msg-right-muted');      
+        var deliveryStatusDiv= $applozic(e.target).closest('.mck-clear').find('.mck-msg-right-muted');  
+        var fileMetaKey = attachmentDiv[0].dataset.filemetakey;    
         if(Kommunicate.internetStatus) {
-            if(!stopUploadIconHidden && uploadIconHidden) {
+            if((!stopUploadIconHidden && uploadIconHidden) || fileMetaKey == "undefined") {
```
ab14bhardwaj: @jithinbalu Why undefined in string ?

jithinbalu: @ab14bhardwaj the value is getting from HTML data attribute that is why undefined is a string

PR #391, opened 2020-02-19
FW-584 Fix alignment problem in mobile browsers with smaller width.

akshatdhabi:
- Fix alignment problem happening in the browser when the mobile width is smaller

## PROD 
image

## FIX
image

### How was the code tested?

- Code was tested on multiple screen widths between 320 - 400px

- Fixed widths were causing issues in smaller widths because they aren't responsive.

### DIFF
--- a/webplugin/template/mck-sidebox.html
+++ b/webplugin/template/mck-sidebox.html
@@ -3,7 +3,7 @@
 		<div id="mck-sidebox-content" class="mck-sidebox-content mck-box-content">
 			<div id="mck-conversation-header" class="mck-conversation-header mck-truncate n-vis"></div>
 			<div class="mck-box-top mck-title">
-				<div class="mck-header-container">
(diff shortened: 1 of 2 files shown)


webplugin/template/mck-sidebox.html
```diff
@@ -3,10 +3,10 @@
 		<div id="mck-sidebox-content" class="mck-sidebox-content mck-box-content">
 			<div id="mck-conversation-header" class="mck-conversation-header mck-truncate n-vis"></div>
 			<div class="mck-box-top mck-title">
-				<div class="blk-lg-10">
+				<div class="mck-header-container">
```
ab14bhardwaj: @akshatdhabi  Please use "Km" instead of "mck" here.
We were using "mck" declarations for Applozic and Kommunicate we are using "km".

PR #392, opened 2020-02-20
PD-296 Allow users to place widget on the left corner of the screen.

akshatdhabi:
- Add support for moving the chat widget to the bottom left of the screen. 

image

### How was the code tested?

- Code was tested by sending the following param in the appsettings API 
`{ 
	"chatWidget": { 
	"moveWidgetToLeft" : true
	}
}`

### DIFF
--- a/webplugin/css/app/mck-sidebox-1.0.css
+++ b/webplugin/css/app/mck-sidebox-1.0.css
@@ -258,7 +258,7 @@
 }
 
 .mck-sidebox .mck-box-title {
-	max-width: 170px;
+	/* max-width: 170px; */
 	font-size: 18px;
 	font-style: normal;
 	font-stretch: normal;
@@ -4433,6 +4433,9 @@ textarea::placeholder {
 	font-stretch: normal;
 	letter-spacing: 0.7px;
 	color: #ffffff;
+}
+div#mck-tab-title.mck-tab-title{
+	text-overflow: ellipsis;
 }
 #mck-conversation-back-btn {
 	-webkit-appearance: none;
--- a/webplugin/css/app/mck-sidebox-1.0.css
+++ b/webplugin/css/app/mck-sidebox-1.0.css
@@ -4772,16 +4772,30 @@ box-shadow: 0 0 4px rgba(0,0,0,.22), 0 4px 8px rgba(0,0,0,.31);
     align-items: center;
 	justify-content: flex-start;
 }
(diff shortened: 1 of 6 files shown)


webplugin/css/app/mck-sidebox-1.0.css
```diff
@@ -4772,16 +4772,30 @@ box-shadow: 0 0 4px rgba(0,0,0,.22), 0 4px 8px rgba(0,0,0,.31);
...
 .mck-msg-preview-visual-indicator-container .mck-close-btn-container, .chat-popup-widget-container .chat-popup-widget-close-btn-container {
 	position: absolute;
     top: 50%;
-    left: -26px;
     display: none;
     transition: all 0.3s ease-in-out;
     transform: translateY(-50%);
     height: 100%;
-    width: 26px;
+	width: 26px;
```
devashishmamgain: formatting

webplugin/js/app/kommunicate-ui.js
```diff
@@ -515,6 +515,9 @@ handleAttachmentIconVisibility : function(enableAttachment, msg, groupReloaded)
             popupTemplateKey === KommunicateConstants.CHAT_POPUP_TEMPLATE.HORIZONTAL ? kommunicateIframe.classList.add('chat-popup-widget-horizontal') : kommunicateIframe.classList.add('chat-popup-widget-vertical');
             kommunicateCommons.modifyClassList( {id : ["launcher-svg-container"]}, "km-animate", "");
             kommunicateCommons.modifyClassList( {id : ["chat-popup-widget-container"]}, "km-animate", "n-vis");
+            var moveWidgetToLeft = kommunicate && kommunicate._globals && kommunicate._globals.widgetSettings && kommunicate._globals.widgetSettings.hasOwnProperty('moveWidgetToLeft') && kommunicate._globals.widgetSettings.moveWidgetToLeft;
```
devashishmamgain: can this be changed to something else, move is like when something is initially positioned somewhere and then moved dynamically. this is more like a position from the beginning.

devashishmamgain: in future if add more positions, it will be difficult to maintain multiple properties, something like below can be used

position: 'bottom-right' (default)
position: 'bottom-left'
position: 'top-right' and so on...

ab14bhardwaj:
@akshatdhabi Add constants for 'left' and 'right'.

PR #418, opened 2020-03-12
disable keyboard popup when a user click on suggestion chips

akshatdhabi:
- Whenever someone was clicking on a suggested reply the keyboard would popup leading to disruption, so disabled the behavior.


webplugin/js/app/km-rich-text-event-handler.js
```diff
@@ -469,8 +469,10 @@ Kommunicate.richMsgEventHandler = {
         console.log("passenger detail submitted");
     },
     processQuickReplies : function(e){
-       var message = e.target.title;
-       var metadata = {};
+        var message = e.target.title;
+        var metadata = {
```
devashishmamgain: @akshatdhabi is this going to be sent to server? if yes, then need to find some other solution, need to figure out a way to achieve this on client side itself

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -3586,7 +3586,8 @@ var KM_ATTACHMENT_V2_SUPPORTED_MIME_TYPES = ["application","text","image"];
                 $mck_msg_sbmt.attr('disabled', false);
                 $applozic('.' + randomId + ' .mck-message-status').removeClass('mck-sent-icon').addClass('mck-pending-icon');
                 mckMessageLayout.addTooltip(randomId);
-                mckMessageLayout.clearMessageField(true);
+                var keyboard = (kommunicateCommons.isObject(messagePxy.metadata) && messagePxy.metadata.hasOwnProperty('keyboard')) ? messagePxy.metadata.keyboard : true ;
```
devashishmamgain: one approach can be to add some specific data type on send button on click of rich message to notify when to not open the keyboard

PR #447, opened 2020-04-03
Disable CSAT button once the rating is submitted.

hardikparashar:
-> Add support to disable CSAT button once the rating is submitted to avoid multiple clicks.

### DIFF
--- a/webplugin/css/app/mck-sidebox-1.0.css
+++ b/webplugin/css/app/mck-sidebox-1.0.css
@@ -5466,12 +5466,12 @@ div#mck-rated:before{
 }
 
 .mck-send-feedback {
-	margin-top: 10px;
+	margin-top: 10px !important;
     padding: 10px;
     width: 100%;
 	border-radius: 4px;
-	color: #fff;
-    font-size: 15px;
+	color: #fff !important;
+    font-size: 15px !important;
 }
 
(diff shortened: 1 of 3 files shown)


webplugin/css/app/mck-sidebox-1.0.css
```diff
@@ -5466,12 +5466,12 @@ div#mck-rated:before{
 }
 
 .mck-send-feedback {
-	margin-top: 10px;
+	margin-top: 10px !important;
```
devashishmamgain, line 5469: please check if using !important is a good practive?
@akshatdhabi any suggestions here?

akshatdhabi: Using `!important` is considered as a bad practice and usually should be done as a last resort as it can cause issues in CSS code down the line. Instead, we can use **CSS Specificity**  to eliminate the need of `!important` almost all the time. 

Please refer: 
https://uxengineer.com/css-specificity-avoid-important-css/

hardikparashar: I was already using an id on the element but the changes are still being overriden by some other styling deifned by the class of the element.

akshatdhabi: The above-mentioned link can easily solve the problem you mentioned.

PR #450, opened 2020-04-06
Need to update chat widget Fullscreen doc

Samir1810:
Need to remove "conversationTitle" and add the Kommunicate chat widget on Pageload doc

https://github.com/Kommunicate-io/Kommunicate-Web-SDK/tree/master/kommunicate-script-example/README.md

-> Add instruction for Kommunicate chat widget on Full screen on Pageload

### How was the code tested?

-> YES

-> Nil

-> Nil

### DIFF
--- a/kommunicate-script-example/README.md
+++ b/kommunicate-script-example/README.md
@@ -42,7 +42,54 @@ Example:
         })(document, window.kommunicate || {});
 
 ```
-> **Note:** Replace the `<APP_ID>` and `<CONVERSATION_TITLE>` with the `appId` and `conversationTitle` copied from the <a href="https://dashboard.kommunicate.io/settings/install" target="_blank">Install Section</a> from the Kommunicate Dashboard.
+> **Note:** Replace the `<APP_ID>` with the `appId` copied from the <a href="https://dashboard.kommunicate.io/settings/install" target="_blank">Install Section</a> from the Kommunicate Dashboard.
+
+
+### Make chat widget fullscreen on pageload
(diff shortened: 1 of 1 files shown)


kommunicate-script-example/README.md
```diff
@@ -42,7 +42,54 @@ Example:
...
+                        element.addEventListener('click', function () {
+                            iframe.classList.add("change-kommunicate-iframe-height");
+                        });
+                    });
+
+                    closeButton.addEventListener('click', function () {
+                        iframe.classList.remove("change-kommunicate-iframe-height");
+                    });
+                    Kommunicate.launchConversation();
+                    KommunicateGlobal.document.getElementById('mck-sidebox-launcher').click()
```
ab14bhardwaj, line 82: @Samir1810 Is the below line needed? I guess your code will work without it also.
KommunicateGlobal.document.getElementById('mck-sidebox-launcher').click()

Samir1810: @ab14bhardwaj  without that line the chat was not opening to full screen. We need to close and open again to show as fullscreen. As mentioned below.

https://postimg.cc/Z0cCKy4Y

devashishmamgain, line 82: please check if  Kommunicate.launchConversation(); is required if we are already using
                    KommunicateGlobal.document.getElementById('mck-sidebox-launcher').click()
are both lines of code required to make it work?

Samir1810: Without that line, it's not opening as fullscreen. Just showing as a button and once clicked, then it's showing full-screen

PR #458, opened 2020-04-21
sending user email to elastic search

Suraj-Kumar:
- Sending user email to elastic search

### How was the code tested?

- 
Screenshot 2020-04-21 at 8 28 43 PM

-

-

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -7516,7 +7516,7 @@ var KM_ATTACHMENT_V2_SUPPORTED_MIME_TYPES = ["application","text","image"];
             _this.updateUser = function(options) {
                 window.Applozic.ALApiService.ajax({
                     type: "POST",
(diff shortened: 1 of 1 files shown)


Suraj-Kumar:
> why to pass it from UI and not do it as default behaviour on server side?
> what will happen to mobile sdks/apps calling existing API.

That's how API is developed. if API behavior needs to be changed that will be a separate story.

devashishmamgain:
> > why to pass it from UI and not do it as default behaviour on server side?
> > what will happen to mobile sdks/apps calling existing API.
> 
> That's how API is developed. if API behavior needs to be changed that will be a separate story.

bug reported says "User email id is getting update in database if "Collect email ID from anonymous users" is enabled but not in elastic search".
No where it is mentioned if there is a bug then pass from UI :)

devashishmamgain:
the way i see it is, there is a bug and the root cause is the API implementation with unnecessary parameter. we should fix the root cause and save all confusion on the code base and avoid creating unnecessary work for all the client sdks, android, iOS, flutter, ionic + chances of more bugs.

PR #464, opened 2020-05-11
Add support for link in card carousel on the top of cards

hardikparashar:
- Fix the issue where URL coming in carousel type rich-messages coming in the google-assistant based payloads was not mapped in the rich-message displayed in chat widget.

### How was the code tested?

- Set a payload in Dialogflow fulfillment with carousel cards.
- Received the message in chat-widget in the form of a card carousel.
- All the items in the carousel are clickable and redirect to the URL in the payload.

Screenshot 2020-05-12 at 1 32 06 PM

### DIFF
--- a/webplugin/js/app/km-richtext-markup-1.0.js
+++ b/webplugin/js/app/km-richtext-markup-1.0.js
@@ -258,8 +258,11 @@ getCarouselTemplate: function() {
 return `<div class="mck-msg-box-rich-text-container km-card-message-container  km-div-slider">
             {{#payload}}
             <div class="km-carousel-card-template">
+            <div class="km-carousel-card-header-container">
+            {{#url}}<a href = {{url}} target="_blank"><span class="km-carousel-url-container"></span></a>{{/url}}
             <div class="km-carousel-card-header {{carouselHeaderClass}}">{{{header}}}</div>
             <div class="km-carousel-card-content-wrapper {{carouselInfoWrapperClass}}">{{{info}}}</div>
+            </div>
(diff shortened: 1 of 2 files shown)


webplugin/js/app/km-richtext-markup-1.0.js
```diff
@@ -258,8 +258,11 @@ getCarouselTemplate: function() {
 return `<div class="mck-msg-box-rich-text-container km-card-message-container  km-div-slider">
             {{#payload}}
             <div class="km-carousel-card-template">
+            <div class="km-carousel-card-header-container">
+            {{#url}}<a href = {{url}} target="_blank"><span class="km-carousel-url-container"></span></a>{{/url}}
```
ab14bhardwaj, line 262: @hardikparashar check with Suraj or Devashish if we will give "openLinkInNewTab" options here for URL click event as we do at other places. If, yes then need to make changes according to that.

devashishmamgain:
Yes if that's possible, let's add it

webplugin/js/app/km-richtext-markup-1.0.js
hardikparashar: The openLinkInNewTab options is not received from the google assistant payload. I checked for that. Only the url is received.

hardikparashar:
@ab14bhardwaj Discussed with @devashishmamgain and @Suraj-Kumar . Since the user can't pass this property from google actions payload we can open the link in a new tab as is the default behaviour of google assistant carousel cards.

PR #475, opened 2020-05-25
TD-826: fix - data is not getting stored into the browser cookie for some websites.

jithinbalu:
- fix: data is not getting stored into the browser cookie

### PR Checklist

### How was the code tested?

- tested in this website https://comtele.com.br/ - new user is not getting created everytime
- When I add "domainKey": "bl" on kommunicateSettings, end-users are same for react.balushome.com and balushome.com websites

- When storing data into cookie we are passing "com.br" as a domain in the cookie URL(document.cookie)
-fix found from here: http://jsfiddle.net/zEwsP/4/

### DIFF
--- a/webplugin/js/app/km-utils.js
+++ b/webplugin/js/app/km-utils.js
@@ -327,10 +327,14 @@ KommunicateUtils = {
         var i = 0;
         var parts = domain.split('.');
         var value = 'km_' + (new Date()).getTime();
+        //check value is added in cookie else continue the iteration
         while (i < (parts.length - 1) && document.cookie.indexOf(value + '=' + value) == -1) {
+            //join the parts of domain
             domain = parts.slice(-1 - (++i)).join('.');
+            //set value in cookie
             document.cookie = value + "=" + value + ";domain=" + domain + ";";
         }
+        //delete value from cookie
         document.cookie = value + "=;expires=Thu, 01 Jan 1970 00:00:01 GMT;domain=" + domain + ";";
         return domain
     },
--- a/webplugin/js/app/km-utils.js
+++ b/webplugin/js/app/km-utils.js
@@ -324,6 +324,7 @@ KommunicateUtils = {
         }
     },
     findCookieDomain: function(domain) {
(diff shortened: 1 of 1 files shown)


webplugin/js/app/km-utils.js
```diff
@@ -324,25 +324,15 @@ KommunicateUtils = {
...
-            }
+        var i = 0;
+        var domain = document.domain;
+        var parts = domain.split('.');
+        var value = 'km_' + (new Date()).getTime();
+        while (i < (parts.length - 1) && document.cookie.indexOf(value + '=' + value) == -1) {
+            domain = parts.slice(-1 - (++i)).join('.');
+            document.cookie = value + "=" + value + ";domain=" + domain + ";";
         }
+        document.cookie = value + "=;expires=Thu, 01 Jan 1970 00:00:01 GMT;domain=" + domain + ";";
```
ab14bhardwaj, line 335: @jithinbalu I don't think we need to do this step.
Tried on jsFiddle, working fine without this also.

jithinbalu: @ab14bhardwaj yes it will work. 
km_{time_stamp} we are storing into the cookie, so this line is using to clear that data. we are not using that for any purpose.

ab14bhardwaj: @jithinbalu Can you add comments for the above code like what's happening inside it.
It would be helpful for others in the future. 
I am approving the PR but please add it before merging it.

ab14bhardwaj: @jithinbalu You can also try this. This is also working and giving proper domain.
It's a small solution. We can use this if it fulfills our use cases.
```
window.location.hostname.split(/[/?#]/)[0];
```
For more information refer below link,
http://scratch99.com/web-development/javascript/how-to-get-the-domain-from-a-url/

jithinbalu: @ab14bhardwaj comments are added.
The solution you have shared doesn't match with the requirement.

Example: https://docs.kommunicate.io/
Query: window.location.hostname.split(/[/?#]/)[0];
output: docs.kommunicate.io
expected output: kommunicate.io

PR #532, opened 2020-08-12
Do not call getDomainFromUrl method multiple times

jithinbalu:
- do not call getDomainFromUrl method multiple times

### PR Checklist

### How was the code tested?

- findCookieDomain method is getting called only once. useful screenshots are attached below.

- getDomainFromUrl method is called multiple places, along with that findCookieDomain method also called. 
- findCookieDomain method tries to find domain by adding "km_{timestamp}" on the cookie and once it is a success, km_{timestamp} be removed from the cookie but chrome always shows deleted cookie in "cookie in use".

**BEFORE** FIX: 
- on screenshot check the "cookie in use" and  browser bottom panel cookies under storage

Screen Shot 2020-08-12 at 3 45 18 PM

- **AFTER** FIX:

Screen Shot 2020-08-12 at 5 05 34 PM

-  findCookieDomain method should be executed only once when the user loads the script.

### DIFF
--- a/webplugin/js/app/km-utils.js
+++ b/webplugin/js/app/km-utils.js
@@ -342,21 +342,17 @@ KommunicateUtils = {
         document.cookie = value + "=;expires=Thu, 01 Jan 1970 00:00:01 GMT;domain=" + domain + ";";
         return domain
     },
-    getDomainFromUrl: function () {
-        return KommunicateUtils.findCookieDomain(document.domain);
-    },
     getSubDomain : function(){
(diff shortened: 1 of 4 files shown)


devashishmamgain:
good findings @jithinbalu 👍

ab14bhardwaj:
@Nitesh94io Please test this issue by today, we will release the chat-plugin tomorrow.
The release is postponed to tomorrow due to the delay that happened today.

PR #541, opened 2020-08-24
Disable chat widget if Allowed Domains is set

shubhamtibra:
- We want to be able to restrict the chat widget from running on domains that are not inside the list of allowed domains specified by the customer. This PR will check the current host from which the chat widget is running and block if it's not in the specified list.

(if running from localhost then the port number must match too)

### PR Checklist

### How was the code tested?

- Ran the modified changes locally and tested the results.

-

-

### DIFF
--- a/webplugin/js/app/mck-app.js
+++ b/webplugin/js/app/mck-app.js
@@ -320,13 +320,11 @@ function ApplozicSidebox() {
             }
 
             // Remove scripts if chatwidget is restricted by domains
-            if (Array.isArray(allowedDomains) && allowedDomains.length && !allowedDomains.some(isSubDomain)){
-                parent.window && parent.window.removeKommunicateScripts();
-                return false;
-            }
+            var isCurrentDomainAllowed = Array.isArray(allowedDomains) && allowedDomains.length && !allowedDomains.some(isSubDomain);
 
             // Remove scripts if disableChatWidget property is enabled
-            if (disableChatWidget) {
+            // or domain restrictions are enabled
+            if (disableChatWidget || isCurrentDomainAllowed) {
                 parent.window && parent.window.removeKommunicateScripts();
                 return false;
             }
--- a/webplugin/js/app/mck-app.js
+++ b/webplugin/js/app/mck-app.js
@@ -311,10 +311,12 @@ function ApplozicSidebox() {
             var disableChatWidget = options.disableChatWidget != null ? options.disableChatWidget : widgetSettings.disableChatWidget; // Give priority to appOptions over API data.
             
             var allowedDomains = widgetSettings.allowedDomains;
-            var host = parent.window.location.host;
+            var hostname = parent.window.location.hostname;
+
+            const isSubDomain = (domain) => hostname.endsWith(domain);
 
             // Remove scripts if chatwidget is restricted by domains
-            if (Array.isArray(allowedDomains) && allowedDomains.length && !allowedDomains.includes(host)){
+            if (Array.isArray(allowedDomains) && allowedDomains.length && !allowedDomains.some(isSubDomain)){
                 parent.window && parent.window.removeKommunicateScripts();
                 return false;
             }
--- a/webplugin/js/app/mck-app.js
+++ b/webplugin/js/app/mck-app.js
@@ -309,7 +309,15 @@ function ApplozicSidebox() {
             var options = applozic._globals;
             var widgetSettings = data.chatWidget;
             var disableChatWidget = options.disableChatWidget != null ? options.disableChatWidget : widgetSettings.disableChatWidget; // Give priority to appOptions over API data.
-            
+            var allowedDomains = widgetSettings.allowedDomains;
+            var host = parent.window.location.host;
+            console.log(host);
+            if (Array.isArray(allowedDomains) && allowedDomains.length){
+              if(!allowedDomains.includes(host)){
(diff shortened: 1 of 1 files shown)


webplugin/js/app/mck-app.js
```diff
@@ -309,7 +309,15 @@ function ApplozicSidebox() {
             var options = applozic._globals;
             var widgetSettings = data.chatWidget;
             var disableChatWidget = options.disableChatWidget != null ? options.disableChatWidget : widgetSettings.disableChatWidget; // Give priority to appOptions over API data.
-            
+            var allowedDomains = widgetSettings.allowedDomains;
+            var host = parent.window.location.host;
+            console.log(host);
```
ab14bhardwaj, line 314: @shubhamtibra Please remove "console.log"

ab14bhardwaj:
@shubhamtibra Please fill all the fields in the PR template provided.
Last two questions are optional as per your understandings.

webplugin/js/app/mck-app.js
```diff
@@ -309,7 +309,15 @@ function ApplozicSidebox() {
...
+            if (Array.isArray(allowedDomains) && allowedDomains.length){
+              if(!allowedDomains.includes(host)){
```
devashishmamgain, line 316: @shubhamtibra for cases when there is no else, then combining all conditions into one single if will work fine

devashishmamgain (changes requested):

webplugin/js/app/mck-app.js
```diff
@@ -310,6 +310,15 @@ function ApplozicSidebox() {
             var widgetSettings = data.chatWidget;
             var disableChatWidget = options.disableChatWidget != null ? options.disableChatWidget : widgetSettings.disableChatWidget; // Give priority to appOptions over API data.
             
+            var allowedDomains = widgetSettings.allowedDomains;
+            var host = parent.window.location.host;
+
+            // Remove scripts if chatwidget is restricted by domains
+            if (Array.isArray(allowedDomains) && allowedDomains.length && !allowedDomains.includes(host)){
```
devashishmamgain, line 317: will it work for sub domains as well, eg: if company have configured lets say epic.com and loads script in their multiple sites such as:
blog.epic.com
www.epic.com
dashboard.epic.com

devashishmamgain (changes requested):

shubhamtibra:
I have made the changes as requested. Also, I am on Ubuntu, so how should it be tested for IE?

webplugin/js/app/mck-app.js
shubhamtibra: Each subdomain will have to be configured.

ab14bhardwaj:
> I have made the changes as requested. Also, I am on Ubuntu, so how should it be tested for IE?
@shubhamtibra 
- Good question, Since you're running your machine with Ubuntu, so while writing the code you can check whether the code is compatible in IE or not. By code, I mean constructors, variable types, etc
You can check it on MDN or https://caniuse.com/.

- I have suggested some changes regarding "writing duplicate code" over slack, please check them also.

webplugin/js/app/mck-app.js
devashishmamgain: is it possible to make this code flexible to cover this use case?

shubhamtibra:
Sure. I'll look into it.

webplugin/js/app/mck-app.js
shubhamtibra: Fixed

devashishmamgain (approved):

webplugin/js/app/mck-app.js
```diff
@@ -310,6 +310,17 @@ function ApplozicSidebox() {
             var widgetSettings = data.chatWidget;
             var disableChatWidget = options.disableChatWidget != null ? options.disableChatWidget : widgetSettings.disableChatWidget; // Give priority to appOptions over API data.
             
+            var allowedDomains = widgetSettings.allowedDomains;
+            var hostname = parent.window.location.hostname;
+
+            const isSubDomain = (domain) => hostname.endsWith(domain);
```
ab14bhardwaj, line 316: @shubhamtibra endsWith property is not supported in IE.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith

devashishmamgain (approved):

ab14bhardwaj (approved):

webplugin/js/app/mck-app.js
```diff
@@ -310,8 +310,21 @@ function ApplozicSidebox() {
...
+
+            // Remove scripts if chatwidget is restricted by domains
+            var isCurrentDomainAllowed = Array.isArray(allowedDomains) && allowedDomains.length && !allowedDomains.some(isSubDomain);
+
             // Remove scripts if disableChatWidget property is enabled
-            if (disableChatWidget) {
+            // or domain restrictions are enabled
+            if (disableChatWidget || isCurrentDomainAllowed) {
                 parent.window && parent.window.removeKommunicateScripts();
                 return false;
```
gargshantnu, line 329: if the current domain is allowed then why are we removing km script?

shubhamtibra: misnamed the variable. Will fix it.

gargshantnu (approved):

devashishmamgain (approved):

ab14bhardwaj (approved):

PR #542, opened 2020-08-31
Add voice output in chat widget

shubhamtibra:
- We want to add the functionality of enabling a new option voiceOutput in the chatWidget settings. When it is true, the browser will convert incoming messages to voice and speak them to the user.

Currently, only text messages, location and attachment messages will be converted to voice.

### PR Checklist

### How was the code tested?

-

-

-


webplugin/js/app/media/media-service.js
```diff
@@ -41,5 +41,38 @@ Kommunicate.mediaService = {
...
+                textToSpeak += message.message;
+
+            }
+        }
+        if (textToSpeak) {
+        var utterance = new SpeechSynthesisUtterance(textToSpeak);
+        
+        utterance.onerror = ev =>{
+            console.log("Error occured in speech synthesis " + ev.error);
+        };
```
ab14bhardwaj, line 73: @shubhamtibra  We don't print logs in our production builds, so you can throw the error rather than logging.

ab14bhardwaj (dismissed):

ab14bhardwaj:
@devashishmamgain Are we thinking of adding this setting in the dashboard or just thinking of giving it as an appOption to add in the install script like "voiceInput"?

devashishmamgain:
> @devashishmamgain Are we thinking of adding this setting in the dashboard or just thinking of giving it as an appOption to add in the install script like "voiceInput"?

We will add it under dashboard settings along with existing voice input feature. Bot can be done together as part of separate issue

devashishmamgain (dismissed):

ab14bhardwaj (dismissed):

PR #545, opened 2020-09-04
Allow chat widget to work on kommunicate.io

shubhamtibra:
- Allow the chat widget to work on kommunicate.io domain even if domain restriction is enabled or chat widget is disabled. 

### PR Checklist

### How was the code tested?

-

-

-

### DIFF
--- a/webplugin/js/app/mck-app.js
+++ b/webplugin/js/app/mck-app.js
@@ -309,12 +309,6 @@ function ApplozicSidebox() {
             var options = applozic._globals;
             var widgetSettings = data.chatWidget;
             var disableChatWidget = options.disableChatWidget != null ? options.disableChatWidget : widgetSettings.disableChatWidget; // Give priority to appOptions over API data.
-            
-            // exclude kommunicate.io from restricted domains for
-            // the chatbot preview feature
-            var kommunicateDomains = [];
-            var kommunicateDomain = "kommunicate.io";
-            kommunicateDomains.push(kommunicateDomain);
 
             var allowedDomains = widgetSettings.allowedDomains;
             var hostname = parent.window.location.hostname.toLowerCase();
@@ -327,7 +321,9 @@ function ApplozicSidebox() {
 
             // Remove scripts if chatwidget is restricted by domains
             var isCurrentDomainDisabled = Array.isArray(allowedDomains) && allowedDomains.length && !allowedDomains.some(isSubDomain);
-            var isCurrentDomainKommunicate = kommunicateDomains.some(isSubDomain);
+            // exclude kommunicate.io from restricted domains for
+            // the chatbot preview feature
+            var isCurrentDomainKommunicate = KommunicateConstants.KOMMUNICATE_DOMAINS.some(isSubDomain);
             // Remove scripts if disableChatWidget property is enabled
             // or domain restrictions are enabled
             if ((disableChatWidget || isCurrentDomainDisabled) && !isCurrentDomainKommunicate) {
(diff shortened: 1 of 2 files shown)


webplugin/js/app/mck-app.js
```diff
@@ -310,6 +310,12 @@ function ApplozicSidebox() {
             var widgetSettings = data.chatWidget;
             var disableChatWidget = options.disableChatWidget != null ? options.disableChatWidget : widgetSettings.disableChatWidget; // Give priority to appOptions over API data.
             
+            // exclude kommunicate.io from restricted domains for
+            // the chatbot preview feature
+            var kommunicateDomains = [];
```
devashishmamgain, line 315: Can we directly write ["kommunicate.io"]

devashishmamgain (changes requested):

webplugin/js/app/mck-app.js
ab14bhardwaj: @shubhamtibra Add this as constant on km-utils files.
ALLOWED_DOMAINS = [ "kommunicate.io"]
In future, we can add other domains if needed in this array only.

webplugin/js/app/mck-app.js
```diff
@@ -321,10 +327,10 @@ function ApplozicSidebox() {
 
             // Remove scripts if chatwidget is restricted by domains
             var isCurrentDomainDisabled = Array.isArray(allowedDomains) && allowedDomains.length && !allowedDomains.some(isSubDomain);
-
+            var isCurrentDomainKommunicate = kommunicateDomains.some(isSubDomain);
             // Remove scripts if disableChatWidget property is enabled
             // or domain restrictions are enabled
-            if (disableChatWidget || isCurrentDomainDisabled) {
+            if ((disableChatWidget || isCurrentDomainDisabled) && !isCurrentDomainKommunicate) {
```
ab14bhardwaj, line 333: @shubhamtibra You can concat two arrays.
1. Array which is coming from widget settings 
2. Array which you have hard coded here.

You won't need an extra variable and check then.
NOTE: spread operator and object.assign doesn't work in IE. So to merge two arrays, use [array concat](https://www.w3schools.com/Jsref/jsref_concat_array.asp) or some other thing which works in IE.

shubhamtibra: >  var isCurrentDomainDisabled = Array.isArray(allowedDomains) && allowedDomains.length && !allowedDomains.some(isSubDomain);

Allowed Domains might be empty array which means the setting is disabled. If we concat the two, we wouldn't be able to check this condition.

ab14bhardwaj: Good point, if the array is not empty then you can contact, otherwise no need to merge the arrays as it will be allowed on every domain.

webplugin/js/app/mck-app.js
shubhamtibra: will do

webplugin/js/app/mck-app.js
shubhamtibra: The condition to check if domain is allowed and domain is kommunicate needs to be different.
Because if the domain is kommunicate we allow in both cases.

shubhamtibra: If i put them in a single array I will get a single result but I need the kommunicate result to allow in disabled settings too.

ab14bhardwaj: Sample code:
```
        var isCurrentDomainDisabled;
        if (Array.isArray(allowedDomains) && allowedDomains.length) {
            allowedDomains = allowedDomains.concat(["kommunciate.io"]); // add your concat logic here 
            isCurrentDomainDisabled = !allowedDomains.some(isSubDomain);
        }
```
If something is failing in the above code, ping me over slack.
Thanks.

devashishmamgain (approved):

PR #569, opened 2020-10-19
naming correction

Kusum-Dhoundiyal:
-

### PR Checklist

### How was the code tested?

-

-

-

### DIFF
--- a/webplugin/js/app/labels/default-labels.js
+++ b/webplugin/js/app/labels/default-labels.js
@@ -163,7 +163,7 @@ Kommunicate.defaultLabels = {
         'attachment': 'You have an attachment.'
     },
     'waiting.queue.message':{
-        'first.Part':'You are currently',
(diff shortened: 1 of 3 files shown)


devashishmamgain (approved):

webplugin/js/app/labels/default-labels.js
```diff
@@ -163,7 +163,7 @@ Kommunicate.defaultLabels = {
         'attachment': 'You have an attachment.'
     },
     'waiting.queue.message':{
-        'first.Part':'You are currently',
+        'first.part':'You are currently',
```
ab14bhardwaj, line 166: @Kusum-Dhoundiyal Do we need to update the same in documentation also?

ab14bhardwaj: @Kusum-Dhoundiyal Can you check this?

PR #571, opened 2020-10-23
Allow file upload with html in their name and Escape the HTML tags when showing in UI

shubhamtibra:
- Allow to upload files with html in their names. While rendering it in the UI, escape the HTML so it shown normally and not rendered as HTML.

### PR Checklist

### How was the code tested?

- Upload a file named `<h1> test <h1>.txt` and `<h1> test <h1>.png`.
- Check that they are uploaded normally
- HTML should not be rendered

Screenshot 2020-10-23 at 12 32 07 PM
Screenshot 2020-10-23 at 12 32 12 PM

-

-

### DIFF
--- a/webplugin/js/app/km-utils.js
+++ b/webplugin/js/app/km-utils.js
@@ -396,7 +396,4 @@ KommunicateUtils = {
             '(\\#[-a-z\\d_]*)?$', 'i'); // fragment locator
         return pattern.test(str);
     },
-    removeHtmlTags: function (text) {
-        return text.replace(/</g, "&lt;").replace(/>/g, "&gt;");
-    }
 }
--- a/webplugin/js/app/km-utils.js
+++ b/webplugin/js/app/km-utils.js
@@ -395,5 +395,8 @@ KommunicateUtils = {
             '(\\?[;&a-z\\d%_.~+=-]*)?'+ // query string
             '(\\#[-a-z\\d_]*)?$', 'i'); // fragment locator
         return pattern.test(str);
+    },
+    removeHtmlTags: function (text) {
+        return text.replace(/</g, "&lt;").replace(/>/g, "&gt;");
     }
 }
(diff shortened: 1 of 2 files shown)


devashishmamgain (dismissed):

webplugin/js/app/km-utils.js
```diff
@@ -395,5 +395,8 @@ KommunicateUtils = {
             '(\\?[;&a-z\\d%_.~+=-]*)?'+ // query string
             '(\\#[-a-z\\d_]*)?$', 'i'); // fragment locator
         return pattern.test(str);
+    },
+    removeHtmlTags: function (text) {
```
ab14bhardwaj, line 399: @shubhamtibra Similar function is already there, you can use that.
kommunicateCommons.formatHtmlTag(data)

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -8795,72 +8795,69 @@ var KM_ATTACHMENT_V2_SUPPORTED_MIME_TYPES = ["application","text","image"];
                     $file_remove.attr("disabled", true);
                     $mck_file_upload.attr("disabled", true);
                     KommunicateUI.hideFileBox(file, $file_box, $mck_file_upload);
-                    if (params.name === $applozic(".mck-file-box." + randomId + " .mck-file-lb a").html()) {
```
ab14bhardwaj, line 8798: @shubhamtibra Any specific reason to remove this check?
```
if (params.name === $applozic(".mck-file-box." + randomId + " .mck-file-lb a").html()) {
```

ab14bhardwaj (changes requested):

webplugin/js/app/mck-sidebox-1.0.js
shubhamtibra: This check is to stop file upload with HTML in their them right?

I thought since we're not rendering the HTML, we can allow upload.

ab14bhardwaj: From my knowledge, I guess we match params.name with the text present inside the element.
We're fetching the element using jquery.

shubhamtibra: Yeah, the upload for files like `<h1> test <h1>.txt` doesn't work with this check.

ab14bhardwaj (approved):

devashishmamgain (approved):

PR #579, opened 2020-11-02
Remove attachment option under settings 

Prochnost55:
- add an option in settings for removing attachment from the widget

### PR Checklist

### How was the code tested?

- PR: https://github.com/Kommunicate-io/Kommunicate-Dashboard/pull/79
attachment parameter in widget is changed

-

-

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -65,7 +65,8 @@ var KM_ATTACHMENT_V2_SUPPORTED_MIME_TYPES = ["application","text","image"];
             'defaultChatDisabledMessage': 'Chat Disabled!'
         },
         voiceInput:false,
-        voiceOutput:false
+        voiceOutput:false,
+        attachment:false
     };
     var message_default_options = {
         'messageType': 5,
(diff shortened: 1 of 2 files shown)


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -65,7 +65,8 @@ var KM_ATTACHMENT_V2_SUPPORTED_MIME_TYPES = ["application","text","image"];
             'defaultChatDisabledMessage': 'Chat Disabled!'
         },
         voiceInput:false,
-        voiceOutput:false
+        voiceOutput:false,
+        attachment:false
```
ab14bhardwaj, line 69: @Prochnost55 We can remove this.
We need to keep this setting "true" by default which is already being done in below line.
var MCK_ATTACHMENT = (typeof appOptions.attachment === "boolean") ? appOptions.attachment : true;

In Dashboard also, if nothing is coming in chatWidget object then treat it as "true".

Prochnost55: Okay, sir, I have done the required changes.

ab14bhardwaj (approved):

PR #588, opened 2020-12-04
add attachment setting support in plugin

Kusum-Dhoundiyal:
-

### PR Checklist

### How was the code tested?

-

-

-


ab14bhardwaj (changes requested):
@Kusum-Dhoundiyal Please add an assignee so that this PR can be reviewed and merged.

ab14bhardwaj (approved):

PR #597, opened 2021-02-02
fix this error [err_cleartext_not_permitted] [VOGO]

Prochnost55:
- To fix an issue raised by VOGO in which following error was shown err_cleartext_not_permitted

### PR Checklist

### How was the code tested?

-

-

- Cleartext is any transmitted or stored information that is not encrypted or meant to be encrypted. When an app communicates with servers using a cleartext network traffic, such as HTTP, it could raise the risk of eavesdropping and tampering of content. Third parties can inject unauthorized data or leak information about the users. That is why developers are encouraged to secure traffic only, such as HTTPS. Starting with Android 9.0 (API level 28), cleartext support is disabled by default. Due to security purposes URL without HTTPS will throw err_cleartext_not_permitted error whenever an application uses it in the Android webview.
There’s an easy solution to fix err_cleartext_not_permitted error and i.e. don’t use insecure URLs. It is recommended to force HTTPs on your websites and remove all the insecure URLs i.e. non-HTTPs from the application.

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -5734,11 +5734,11 @@ var KM_ATTACHMENT_V2_SUPPORTED_MIME_TYPES = ["application","text","image"];
                   try {
                       var geoLoc = $applozic.parseJSON(msg.message);
                       if (geoLoc.lat && geoLoc.lon) {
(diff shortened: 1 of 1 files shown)


devashishmamgain (commented):
will it have any problem with customers whose website is in HTTP?

devashishmamgain (approved):

Prochnost55:
> will it have any problem with customers whose website is in HTTP?

No sir, I don't think there should be any problem, as upon clicking the location link the browser opens a new window and load google maps with that particular location.

ab14bhardwaj (approved):

PR #611, opened 2021-02-10
TD-1530 Add option in the chat widget to turn off text to speech at end-user side

Prochnost55:
- Create a dropdown option in the chat widget to toggle the voice output.

### PR Checklist

### How was the code tested?

- Enable the Text to speech setting from the dashboard.
- Click on menu button to get the option to toggle the voice output
Screenshot 2021-02-10 at 3 09 10 PM

-

-


webplugin/js/app/media/media-service.js
```diff
@@ -45,13 +45,13 @@ Kommunicate.mediaService = {
     voiceOutputIncomingMessage: function (message) {
         // get appoptions
         var appOptions = KommunicateUtils.getDataFromKmSession("appOptions") || applozic._globals;
-
+        var userOverrideVoiceOutput = KommunicateUtils.getItemFromLocalStorage("USER_OVERRIDE").VOICE_OUTPUT_ENABLED;
```
devashishmamgain, line 48: is USER_OVERRIDE guaranteed to be always present? if not then this will give undefined error

devashishmamgain (changes requested):

webplugin/js/app/media/media-service.js
Prochnost55: dependency on local storage is removed now

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -9907,13 +9926,15 @@ var KM_ATTACHMENT_V2_SUPPORTED_MIME_TYPES = ["application","text","image"];
                             };
                         };
                         if (messageType === "APPLOZIC_01" || messageType === "MESSAGE_RECEIVED") {
-                            var messageFeed = mckMessageLayout.getMessageFeed(message);
-                            Kommunicate.KmEventHandler.onMessageReceived(message);
+                            // var messageFeed = mckMessageLayout.getMessageFeed(message);
+                            var messageCopy = JSON.parse(JSON.stringify(message));
```
devashishmamgain, line 9930: is this equivalent to mckMessageLayout.getMessageFeed(message)?
above function sets some values for file attachment URLs etc, please check if those are equivalent

devashishmamgain (changes requested):
is this setting specific to per message?
not able to figure out why it is tied to a single message property and not as global setting for that user

webplugin/js/app/mck-sidebox-1.0.js
Prochnost55:
```
var messageFeed = mckMessageLayout.getMessageFeed(message);
```
This line was not getting used as the code below was commented. So that's why I commented the above line.

Prochnost55:
> is this setting specific to per message?
> not able to figure out why it is tied to a single message property and not as global setting for that user

Regarding, the above query. The settings are on user level and not message, we're just passing a variable inside message object which is not accessible outside the mck-sidebox-1.0.js file. So inside our event-handler file, now we can check for that value and make a decision whether to allow voice output or not.

devashishmamgain (dismissed):

devashishmamgain:
@Prochnost55 remove the comment code

ab14bhardwaj (approved):

devashishmamgain (approved):

ab14bhardwaj (approved):

PR #639, opened 2021-03-18
Chat widget was not loading in the IE

Prochnost55:
- Fix the issue where the chat widget was not loading in the IE.

### PR Checklist

### How was the code tested?

- get the branch deployed on widget-test
- load the script in IE

-

- backticks and arrow functions were used in the code which was throwing error in ie


gargshantnu (approved):

shubhamtibra (approved):

shubhamtibra:
`AVAILABLE_VOICES_FOR_TTS.forEach((voice) => {`

This execution should never reach this line in IE since we have checks before `"speechSynthesis" in window`.

shubhamtibra:
Maybe the bug was because of the backticks only?

PR #654, opened 2021-04-12
FW 2013 Confusion on kommunicate and Kommunicate in web chat plugin 

Ojashwa:
**- Remove confusion on kommunicate and Kommunicate in the web chat plugin.**

### PR Checklist

### How was the code tested?

- **Run Kommunicate script in the console and check for properties of both objects (Kommunicate and kommunicate).**
---
image

-

-

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -81,12 +81,8 @@ var userOverride = {
     };
     $applozic.fn.applozic = function (appOptions, params, callback) {
         var $mck_sidebox = $applozic('#mck-sidebox');
-        if (
-            Object.keys(kommunicate).length !== Object.keys(Kommunicate).length
-        ) {
-            $applozic.extend(Kommunicate, kommunicate);
-            $applozic.extend(kommunicate, Kommunicate);
-        }
+        // every time this line will overwrite the kommunicate object properties
+        $applozic.extend(kommunicate, Kommunicate);
         if ($applozic.type(appOptions) === 'object') {
             // storing custom appOptions into session Storage.
             KommunicateUtils.storeDataIntoKmSession('appOptions', appOptions);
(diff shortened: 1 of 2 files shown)


shubhamtibra:
The current structure for the variables are:
1. 
Screenshot 2021-04-12 at 2 42 16 PM
2. 
Screenshot 2021-04-12 at 2 43 05 PM

If both variable need to mean the same thing, then they need to be synchronized when new things are added in future. @devashishmamgain Any help how that can be achieved?

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -81,6 +81,12 @@ var userOverride = {
     };
     $applozic.fn.applozic = function (appOptions, params, callback) {
         var $mck_sidebox = $applozic('#mck-sidebox');
+        if (
+            Object.keys(kommunicate).length !== Object.keys(Kommunicate).length
+        ) {
+            $applozic.extend(Kommunicate, kommunicate);
+            $applozic.extend(kommunicate, Kommunicate);
+        }
```
shubhamtibra, line 89: We need to make sure none of the variables are modified after this line is executed. Otherwise there will be a mismatch.

shubhamtibra (commented):
Is this a good idea? The `kommunicate` variable is only storing data, so the distinction between them is clear. Also I think  `kommunicate` isn't used externally by the customers, right?.
@devashishmamgain

devashishmamgain:
> Is this a good idea? The `kommunicate` variable is only storing data, so the distinction between them is clear. Also, `kommunicate` isn't used externally by the customers, right?. So it shouldn't be a problem?
> @devashishmamgain

the confusion customers are facing is that some places in docs it is kommunicate, some places it is Kommunicate and customers are using it interchangeably and reporting issue that the function mentioned in docs is not working (they are using kommunicate in place of Kommunicate).

keeping a track of variable modification will be difficult, need to figure out some way that both are variables are ultimately same.

shubhamtibra (approved):

Rudranand (approved):

devashishmamgain (approved):

PR #666, opened 2021-05-07
Merge development into plugin-6.7

Prochnost55:
-

### PR Checklist

### How was the code tested?

-

-

-


Prochnost55:
need to simplify the release process on widget side

PR #673, opened 2021-05-13
FW-2226 Make improvements in the file encryption logic in chat widget

Prochnost55:
- The encrypted images should distinguishable from non-encrypted images.
- Both encrypted and non encrypted images should open.
- The videos should be handled 

### PR Checklist

### How was the code tested?

- open chat widget using plugin-6.6.4 or lower upload image A.
- open chat widget using this branch upload image B
- upload a video from this branch 
- all the three media should be visible in the chat widget

- 

-

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -7953,6 +7953,13 @@ var userOverride = {
                     ? '<img src="' + topicLink + '">'
                     : '<span class="mck-icon-no-image"></span>';
             };
+            _this.isFileEncryptedImage = function (fileMeta){
+                return (
+                    fileMeta && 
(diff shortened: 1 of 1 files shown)


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -14438,6 +14460,10 @@ var userOverride = {
                             $file_remove.trigger('click');
                         }
                     });
+                    var ATTACHMENT_UPLOAD_URL = '/rest/ws/upload/image';
```
shubhamtibra, line 14463: Can we keep the constant definition at the same place as before? And add just the required parameter here without editing constant?

shubhamtibra (approved):

Rudranand (approved):

PR #681, opened 2021-05-31
FW-2276 Use KM API URL for bot detail | web plugin

Ojashwa:
-

### PR Checklist

### How was the code tested?

- **use installation script and check for the API details in network panel**
---
image
---
image

 

-

-

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -12694,20 +12694,16 @@ var userOverride = {
                 return conversationDetail;
             };
             _this.checkBotDetail = function (userId) {
-                var authKey = JSON.parse(
-                    decodeURIComponent(
-                        escape(
-                            window.atob(sessionStorage.getItem('chatheaders'))
-                        )
-                    )
-                ).authToken;
                 mckUtils.ajax({
                     url:
                         Kommunicate.getBaseUrl() +
                         '/rest/ws/botdetails/' +
                         userId,
                     type: 'get',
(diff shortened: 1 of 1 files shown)


devashishmamgain (approved):

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -12695,14 +12694,20 @@ var userOverride = {
                 return conversationDetail;
             };
             _this.checkBotDetail = function (userId) {
-                window.Applozic.ALApiService.ajax({
+                var authKey = JSON.parse(
```
devashishmamgain, line 12697: Is this the same which is used in any other api?

Ojashwa: yes.

devashishmamgain: can you share any code reference for that

Ojashwa: how the headers are being set?

devashishmamgain: I mean any other place from where auth header is fetched from storage

Rudranand (dismissed):

webplugin/js/app/mck-sidebox-1.0.js
Ojashwa: Ratified

devashishmamgain (approved):

Rudranand (approved):

PR #690, opened 2021-06-09
use secure route for feedback API

gargshantnu:
## What do you want to achieve?
- use the secure route for feedback API

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -1955,7 +1955,7 @@ var userOverride = {
             var $mck_tab_individual = $applozic('#mck-tab-individual');
             var MCK_IDLE_TIME_COUNTER = MCK_IDLE_TIME_LIMIT;
             var INITIALIZE_APP_URL = '/v2/tab/initialize.page';
-            var FEEDBACK_UPDATE_URL = '/feedback/v2';
+            var FEEDBACK_UPDATE_URL = '/rest/ws/feedback/v2/v2';
             _this.getLauncherHtml = function (isAnonymousChat) {
                 var defaultHtml = kmCustomTheme.customSideboxWidget();
                 var CHAT_CLOSE_BUTTON = `<div id="km-popup-close-button" aria-label="Close" role="button" class="km-custom-widget-background-color">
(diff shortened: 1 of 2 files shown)


devashishmamgain (approved):

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -1955,7 +1955,7 @@ var userOverride = {
             var $mck_tab_individual = $applozic('#mck-tab-individual');
             var MCK_IDLE_TIME_COUNTER = MCK_IDLE_TIME_LIMIT;
             var INITIALIZE_APP_URL = '/v2/tab/initialize.page';
-            var FEEDBACK_UPDATE_URL = '/feedback/v2';
+            var FEEDBACK_UPDATE_URL = '/rest/ws/feedback/v2/v2';
```
devashishmamgain, line 1958: is v2 extra here?

devashishmamgain (changes requested):

webplugin/js/app/mck-sidebox-1.0.js
gargshantnu: yes, in the backend new secure route was `/rest/ws/feedback/v2`
and this API earlier was `/feedback/v2`
so when we replaced `/feedback` with `/rest/ws/feedback/v2`, it got mixed up. and 2 `v2` got added.
once all the old routes are refactored from all SDKs, we can correct this.

devashishmamgain: or add support for v3 and use the same in this PR and add support for both v3 and /v2 and /v2/v2

gargshantnu: means in backend add support for `/v3/v2` and use that here?

devashishmamgain: no i mean, we need to avoid v2/v2, and add support for /v3 as well [just v3 and only once, no duplicate things]

and keep things simple and understandable

gargshantnu: okay. but for that, we need to do backend changes as well. is that okay? i mean, it will then create a dependency of deployment order.

devashishmamgain: ok then leave it as same

devashishmamgain (approved):

Prochnost55 (approved):

PR #691, opened 2021-06-14
FW-2376 User is able to reply after conversation got resolved [VOGO][PROD]

Prochnost55:
- The end user should not be able to reply once conv is resolved

### PR Checklist

### How was the code tested?

- create a conversation
- assign to human agent
- use https://apps-test.applozic.com/rest/ws/group/status/change?groupId=433196&status=2 api to change the status
- upon success response the conv should get resolved

-

- the resp coming in onMessage() is having type 'APPLOZIC_02' and for this case the showClosedConversationBanner() was getting false as argument


Prochnost55:
There was some confusion regarding headers

PR #626, opened 2021-02-24
FW-1870-Show group specific data in conversation section

Ojashwa:
### What do you want to achieve?
-

### PR Checklist

- [x] I have tested it locally and all functionalities are working fine.
- [ ] I have compared it with mocks and all design elements are the same.
- [ ] I have tested it in IE Browser.

### How was the code tested?

- **Start a conversation with installation script.**

image
----
- **Select the JavaScript context to _` Kommunicate widget iframe `_  as shown in the below image.**

image
----
- **After a conversation is get started try to send a metadata object in Kommunicate.updateConversationMetadata().**

image
----

- **Metadata can be shown here.**

image
-

- **Empty state for conversation info.**

image

- https://showgroupspecificdata.netlify.app

### What new thing you came across while writing this code? 
-

### In case you fixed a bug then please describe the root cause of it? 
-

NOTE: Make sure you're comparing your branch with the correct base branch

PR #702, opened 2021-07-01
Fix object check error in updateConversationMetadata function [Prod issue]

Ojashwa:
(#626)

### DIFF
--- a/webplugin/js/app/kommunicate.js
+++ b/webplugin/js/app/kommunicate.js
@@ -124,8 +124,10 @@ $applozic.extend(true, Kommunicate, {
     updateConversationMetadata: function (conversationMetadata) {
         if (conversationMetadata) {
             if (
-                kommunicateCommons.isObject(conversationMetadata) &&
-                kommunicateCommons.isObject(conversationMetadata.metadata) &&
+                Object.prototype.toString.call(conversationMetadata) ==
+                    '[object Object]' &&
+                Object.prototype.toString.call(conversationMetadata.metadata) ==
+                    '[object Object]' &&
                 conversationMetadata.groupId &&
                 conversationMetadata.metadata
             ) {


webplugin/js/app/kommunicate.js
```diff
@@ -124,8 +124,10 @@ $applozic.extend(true, Kommunicate, {
     updateConversationMetadata: function (conversationMetadata) {
         if (conversationMetadata) {
             if (
-                kommunicateCommons.isObject(conversationMetadata) &&
-                kommunicateCommons.isObject(conversationMetadata.metadata) &&
+                Object.prototype.toString.call(conversationMetadata) ==
```
devashishmamgain, line 127: do we need this? or we just need to check if conversationMedta is a JSON?

devashishmamgain (changes requested):

webplugin/js/app/kommunicate.js
Ojashwa: `conversationMetadata` has to be an object and `conversationMetadata.metadata` can be an object or in JSON format.
This is to prevent the pass an array or string for `conversationMetadata.metadata` property.

devashishmamgain: this will work in IE?

also can you attach same sample codes that customers will use

Ojashwa: Sample code after patching: https://jsfiddle.net/m3kwbq49/
For IE: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toString#browser_compatibility

devashishmamgain: jsfiddle is showing error, i guess that evn doesn't have this fix deployed on it?

devashishmamgain (approved):

ChannuBS (approved):

Prochnost55 (approved):

Ojashwa:
created another branch for the same issue (#703)

PR #745, opened 2021-08-04
FW-2531 Text To Speech not working in Opera Browser, Chrome and Edge.

Ojashwa:
- Execute TTS in Opera Browser, Chrome and Edge.

### PR Checklist

### How was the code tested?

- Check TTS for multiple responses and long text in Opera Browser, Chrome and Edge.

-

-

### DIFF
--- a/webplugin/js/app/media/media-service.js
+++ b/webplugin/js/app/media/media-service.js
@@ -51,9 +51,15 @@ Kommunicate.mediaService = {
     },
     voiceOutputIncomingMessage: function (message) {
         // get appoptions
+        var timeOut;
         var appOptions =
             KommunicateUtils.getDataFromKmSession('appOptions') ||
             applozic._globals;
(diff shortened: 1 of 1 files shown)


Prochnost55 (approved):

webplugin/js/app/media/media-service.js
```diff
@@ -120,8 +126,14 @@ Kommunicate.mediaService = {
                         );
                     }
                 };
-                isChrome && speechSynthesis.cancel();
+                if (isChrome) {
```
shubhamtibra, line 129: Do we need to add Opera and Edge check?

Ojashwa: This check will work on Opea and Edge as well. 
Verifed that by testing in the respective browser.

shubhamtibra (approved):

PR #771, opened 2021-09-10
FW-2663 FAQ Search not Working as expected.

Prochnost55:
- fix faq search issues

### PR Checklist

### How was the code tested?

- search for any string in faq search box

-

-

### DIFF
--- a/webplugin/js/app/kommunicate-ui.js
+++ b/webplugin/js/app/kommunicate-ui.js
@@ -319,9 +319,6 @@ KommunicateUI = {
             return false;
         }
     },
-    showCategoryList: function(){
-
-    },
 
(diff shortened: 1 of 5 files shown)


ChannuBS (dismissed):

webplugin/js/app/kommunicate-ui.js
```diff
@@ -718,12 +761,16 @@ KommunicateUI = {
         });
     },
     searchFaqs: function (data) {
+        var searchFilter = {
+            appId: data.appId,
```
Ojashwa, line 765: what if data is null?

Ojashwa (approved):

ChannuBS (approved):

PR #775, opened 2021-09-16
FW-2690 Replace cdn.applozic.com.com with KM CDN URL in km chat widget 

Ojashwa:
- Replace `cdn.applozic.com.com`  with `cdn.kommunicate.io` in the chat widget.

### PR Checklist

### How was the code tested?

- Check for basic functionality in the chat widget.

-

-


devashishmamgain (changes requested):
what's the reason for this? why to remove from CDN. infact we should be adding more and more files to CDN to improve the performance.

@Ojashwa @Prochnost55

devashishmamgain:
move it to Kommunicate CDN

devashishmamgain (approved):

Prochnost55 (approved):

PR #803, opened 2021-11-24
FW-3136 Encrypt all the attachment sent from widget

Prochnost55:
- all the media attachments should be encrypted

### PR Checklist

### How was the code tested?

- go to dashboard, send various types of attachments. All of these attachments should be properly displayed in chat widget.
- go to chat widget send different types of attachments. these attachments should be accessible in dashboard.
- repeat same with dialogflow bot

-

-


ChannuBS (approved):

devashishmamgain:
@Prochnost55 please cover the testing for Dialogflow and other bot attachment send and receive too.

gargshantnu (approved):

PR #815, opened 2021-12-22
[TD-2356] Added support for File-sending-event Handling

raj5036:
- Send/receive files in Zendesk agent to user chat.

### PR Checklist
[
- [ ] I have tested it in IE Browser

### How was the code tested?

-  Refer https://github.com/Kommunicate-io/Kommunicate-Server/pull/388

-  1. Zendesk integration process/steps
-  2. Different usages of Zopim api

-


gargshantnu:
@raj5036 you need to file PR for `development` branch. not for `master`

PR #816, opened 2021-12-23
FW-3132 A banner should be visible if socket connection drops

Prochnost55:
- banner should be visible if socket reconnect fail count is > 2
Screenshot 2021-12-23 at 12 31 26 PM

### PR Checklist

### How was the code tested?

- 
Screenshot 2021-12-23 at 12 33 09 PM

-

-

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -618,6 +618,7 @@ var userOverride = {
             WIDGET_SETTINGS.hasOwnProperty('position')
                 ? WIDGET_SETTINGS.position
                 : KommunicateConstants.POSITION.RIGHT;
+        var SOCKET_RECONNECT_FAIL_COUNT = 0;
         window.Applozic.SOCKET_DISCONNECT_PROCEDURE = {
             SOCKET_DISCONNECT_TIMER_VALUE: 120000, // 2 minutes : 120000 milliSeconds
             DISCONNECTED: false,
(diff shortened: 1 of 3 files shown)


devashishmamgain (changes requested):
is it something which will be visible in debug mode only?
or its visible to everyone?
how will a non -tech end user understand what is "socket"?

gargshantnu (dismissed):

ChannuBS (approved):

Prochnost55:
> is it something which will be visible in debug mode only? or its visible to everyone? how will a non -tech end user understand what is "socket"?

asked from rajeesh 
He suggested to update it to “Conversations will not be refreshed in real time due to network issue”

gargshantnu (approved):

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -716,6 +717,12 @@ var userOverride = {
                 if (navigator.onLine) {
                     window.Applozic.ALSocket.reconnect();
                 }
+                SOCKET_RECONNECT_FAIL_COUNT++;
```
devashishmamgain, line 720: if the user is offline, then also this message will be displayed along with other offline message? or only 1 will be displayed?

devashishmamgain (changes requested):

webplugin/js/app/mck-sidebox-1.0.js
Prochnost55: It will only display one message as shown in the image 
Screenshot 2021-12-27 at 9 32 25 AM

devashishmamgain (approved):

gargshantnu (approved):

ChannuBS (approved):

PR #820, opened 2021-12-31
FW-3229 The widget is initialising with empty header

Prochnost55:
- added a check to prevent undefined from getting passed to JSON.parse().

### PR Checklist

### How was the code tested?

- Please test the code with an website env where
1. the session storage is not available initially.
2. kommunicate is not present in local storage.

-

-

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -7063,7 +7063,7 @@ var userOverride = {
                 isAgentOffline
             ) {
                 var userSession = KommunicateUtils.isSessionStorageAvailable()
-                    ? JSON.parse(sessionStorage.kommunicate)
+                    ? (sessionStorage.kommunicate ? JSON.parse(sessionStorage.kommunicate) : {})
                     : {};
                 var languageCode =
                     userSession &&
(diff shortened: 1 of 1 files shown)


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -7063,7 +7063,7 @@ var userOverride = {
                 isAgentOffline
             ) {
                 var userSession = KommunicateUtils.isSessionStorageAvailable()
-                    ? JSON.parse(sessionStorage.kommunicate)
+                    ? (sessionStorage.kommunicate ? JSON.parse(sessionStorage.kommunicate) : {})
```
gargshantnu, line 7066: `KommunicateUtils.isSessionStorageAvailable()` vs `sessionStorage.kommunicate`
whats the difference?

Prochnost55: sessionStorage is available but it doesn't have `kommunicate` in storage
thus `kommunicate` becomes undefined and `JSON.parse(undefined)` gives error

gargshantnu: any ideas why its not present?

Prochnost55: The customers' website was using some tool to ask user permissions related to cookie and storage
so when the user visits for the 1st time there was no data in storage and also no permission to write data.

gargshantnu, line 7066: can we move the new check to outer level? like combine both the checks as it seems similar

ChannuBS (approved):

gargshantnu (approved):

PR #854, opened 2022-03-31
[FF-92] PreChat login email validation.

ShubhamAgarwalKommunicate:
 User Providing a vaild email address should be allowed to login to the Widget.

### PR Checklist

### How was the code tested?

 Checked by trying to login to the widget with some Invalid email address.

 Got to know the code flow of Kommunicate-Web-SDK repo

 NA

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -3109,7 +3109,19 @@ var userOverride = {
                         );
                         kmChatInput.setAttribute(
                             'pattern',
-                            /^([a-zA-Z0-9_\-\.]+)@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.)|(([a-zA-Z0-9\-]+\.)+))([a-zA-Z]{2,4}|[0-9]{1,3})(\]?)$/
+                            "^[a-zA-Z0-9_+&*-]+(?:\\.[a-zA-Z0-9_+&*-]+)*@(?:[a-zA-Z0-9-]+\\.)+[a-zA-Z]{2,7}$"
+                        );
+                        kmChatInput.setAttribute(
+                            'title',
+                            ''
+                        );
+                        kmChatInput.setAttribute(
+                            'oninvalid',
+                            "setCustomValidity('Please enter a valid email address')"
+                        );
+                        kmChatInput.setAttribute(
+                            'oninput',
+                            "setCustomValidity('')"
                         );
                     }
                     $applozic(kmChatInputDiv).append(kmChatInput);
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -3117,7 +3117,7 @@ var userOverride = {
                         );
                         kmChatInput.setAttribute(
                             'oninvalid',
-                            "setCustomValidity('Please enter a valid email address')"
+                            `setCustomValidity("${MCK_LABELS['lead.collection'].errorEmail}")`
                         );
                         kmChatInput.setAttribute(
                             'oninput',
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -4722,14 +4722,8 @@ var userOverride = {
                                 _this.phoneNumberValidation
                             );
                     }
-                    function ValidateEmail(email) {
-                        var regexp = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
-                        if(regexp.test(String(email).toLowerCase())) return true;
-                        alert("You have entered an invalid email address!");
-                        return false;
-                    }
                     if (email) {
-                        if(!ValidateEmail(email)) return false;
+                        if (!KommunicateUI.validateEmail(email))  return false;
                         userId = email;
                         KommunicateUtils.setCookie({
                             name:
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -4722,7 +4722,14 @@ var userOverride = {
                                 _this.phoneNumberValidation
                             );
                     }
+                    function ValidateEmail(email) {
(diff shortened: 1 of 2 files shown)


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -4722,7 +4722,14 @@ var userOverride = {
                                 _this.phoneNumberValidation
                             );
                     }
+                    function ValidateEmail(email) {
+                        var regexp = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
```
devashishmamgain, line 4726: this same thing is not present in existing code base?

devashishmamgain (changes requested):

webplugin/js/app/mck-sidebox-1.0.js
ShubhamAgarwalKommunicate: Yes the same function was present, I have updated it.

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -4723,6 +4723,7 @@ var userOverride = {
                             );
                     }
                     if (email) {
+                        if (!KommunicateUI.validateEmail(email))  return false;
```
devashishmamgain, line 4726: add { even if it is one line

devashishmamgain (changes requested):

webplugin/js/app/mck-sidebox-1.0.js
devashishmamgain: user will see any error or something if it fails?

devashishmamgain (changes requested):

webplugin/js/app/mck-sidebox-1.0.js
ShubhamAgarwalKommunicate: As of now no error will be promoted, is just the submit wouldn't work. We can add an 'alert' message?

devashishmamgain: yes, tagging Rajeesh in Jira

ShubhamAgarwalKommunicate: Now Html will validate the email, so the same error message format will be shown, have also kept the ValidateEmail() function in case someone tries to remove the validation from the Element section of inspect.

devashishmamgain (dismissed):

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -4723,6 +4727,9 @@ var userOverride = {
                             );
                     }
                     if (email) {
+                        if (!KommunicateUI.validateEmail(email)){
+                            return false;
+                        }
```
Prochnost55, line 4732: If the above regex is correct, then email is always going to be valid. If not, then the end user will see the following message.

Screenshot 2022-04-01 at 10 32 42 AM

So, I wanted to know is there any possibility that the input pattern validation passed but the `validateEmail()` returns false?

devashishmamgain: "Please match the format requested."
where exactly have we requested a format from the user?

ShubhamAgarwalKommunicate: @Prochnost55 While testing I tried removing the regex pattern from the Element section of inspect and it passes the validation so have kept validateEmail() for check in this type of case.
But in general if input pattern validation passes then their is no case that validateEmail() returns false.

ShubhamAgarwalKommunicate: @devashishmamgain Html validates in two way one of it is the default validation in this a proper error message is shown.
loginError
But it do not validates all types of email such as "something@something" so for this have added a regex pattern to validate, this will give error message like "Please match the format requested."

As the coloum itself says to enter email address so is it known that the format should be a valid email id.

As it is validating by html in browser we cant change the error message, for that we need to remove validation from html and add our at our code.

devashishmamgain: this line will be confusing, is there a way to customize the message?

ShubhamAgarwalKommunicate: Yes we can pass a customize message, it will be a static message which will be shown in case of wrong email.
loginerror
We can pass "Please enter a valid email address !!".

devashishmamgain: use following:
Please enter a valid email address

Also, add it under MCK_LABELS
https://docs.kommunicate.io/docs/web-localization

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -3107,6 +3107,22 @@ var userOverride = {
...
+                            'pattern',
+                            "^[a-zA-Z0-9_+&*-]+(?:\\.[a-zA-Z0-9_+&*-]+)*@(?:[a-zA-Z0-9-]+\\.)+[a-zA-Z]{2,7}$"
+                        );
+                        kmChatInput.setAttribute(
+                            'title',
+                            ''
+                        );
+                        kmChatInput.setAttribute(
+                            'oninvalid',
+                            "setCustomValidity('Please enter a valid email address')"
```
devashishmamgain, line 3120: use MCK_LABELS
https://docs.kommunicate.io/docs/web-localization

devashishmamgain (changes requested):

webplugin/js/app/mck-sidebox-1.0.js
ShubhamAgarwalKommunicate: Does this doc also needs to be updated?
have updated in the code.

devashishmamgain (dismissed):

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -3107,6 +3107,22 @@ var userOverride = {
...
+                            'pattern',
+                            "^[a-zA-Z0-9_+&*-]+(?:\\.[a-zA-Z0-9_+&*-]+)*@(?:[a-zA-Z0-9-]+\\.)+[a-zA-Z]{2,7}$"
+                        );
+                        kmChatInput.setAttribute(
+                            'title',
+                            ''
+                        );
+                        kmChatInput.setAttribute(
+                            'oninvalid',
+                            `setCustomValidity("${MCK_LABELS['lead.collection'].errorEmail}")`
```
Prochnost55, line 3120: can you please test this in IE?
It will break in IE as string template is not supported

Prochnost55 (changes requested):

webplugin/js/app/mck-sidebox-1.0.js
ShubhamAgarwalKommunicate: Yes I tested string template is not supported in IE, have changed the code accordingly.

Prochnost55 (approved):

PR #857, opened 2022-04-06
[TD-2538] Fixed Zopim Chat transcript bug

raj5036:
- Fix Zopim Chat transcript bug where sometimes Agent's name was coming instead of Bot's name.

### PR Checklist

### How was the code tested?

- Tested locally. Changes are working as expected.

image

### DIFF
--- a/webplugin/js/app/zendesk-chat-service.js
+++ b/webplugin/js/app/zendesk-chat-service.js
@@ -130,6 +130,7 @@ function ZendeskChatService() {
 
                 var messageListDetails = result.message;
                 var userDetails = result.userDetails;
+                console.log("userDetails are ",userDetails);
 
                 var userId = kommunicate._globals.userId;
 
@@ -146,8 +147,7 @@ function ZendeskChatService() {
                         username = "User";
                     } else {
                         //Get bot name
-                        var botId = currentMessageDetail.to;
-                        username = _this.getBotNameById(botId, userDetails);
+                        username = _this.getBotNameById(currentMessageDetail.to, userDetails);
                     }
 
                     var message = _this.getMessageForTranscript(currentMessageDetail);
--- a/webplugin/js/app/zendesk-chat-service.js
+++ b/webplugin/js/app/zendesk-chat-service.js
@@ -129,6 +129,7 @@ function ZendeskChatService() {
                 }
 
                 var messageListDetails = result.message;
+                var userDetails = result.userDetails;
 
                 var userId = kommunicate._globals.userId;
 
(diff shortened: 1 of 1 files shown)


webplugin/js/app/zendesk-chat-service.js
```diff
@@ -166,6 +169,16 @@ function ZendeskChatService() {
         }
     };
 
+    _this.getBotNameById = function(botId, userDetails) {
+        for (var i = 0; i<userDetails.length; i++) {
+            var currentUserDetail = userDetails[i];
+            if (currentUserDetail.userId == botId) {
+                return currentUserDetail.userName;
+            }
+        }
```
gargshantnu, line 178: 1. can be achieved by simple .find().
2. does not seem efficient, can you think of some more optimised approach

webplugin/js/app/zendesk-chat-service.js
```diff
@@ -144,12 +145,14 @@ function ZendeskChatService() {
                     if (currentMessageDetail.to === userId) {
                         username = "User";
                     } else {
-                        username = currentGroupData.displayName;
+                        //Get bot name
+                        var botId = currentMessageDetail.to;
+                        username = _this.getBotNameById(botId, userDetails);
```
gargshantnu, line 150: botId variable can be removed and directly the value can be used

webplugin/js/app/zendesk-chat-service.js
raj5036: Didn't use find because its not supported on IE.  Need to think of something else then👍
image

webplugin/js/app/zendesk-chat-service.js
raj5036: Ok👍

devashishmamgain (approved):

webplugin/js/app/zendesk-chat-service.js
devashishmamgain: we maintain some map also MCK_USER_DETAIL_MAP or something, see if that can be used

devashishmamgain (changes requested):

webplugin/js/app/zendesk-chat-service.js
raj5036: Done.

webplugin/js/app/zendesk-chat-service.js
raj5036: What I have found is this `userDetails` array generally consists of 4-5 users.

1. The user that is using the chat-widget.
2. The assigned bot for the conversation.
3. The assigned Agent after handoff.
4. And this `suraj++13@applozic.com` is present in every group on that particular account. I guess because this email is added as an Admin or something.

image

So this should not cause performance issues. @devashishmamgain @gargshantnu

devashishmamgain (approved):

Prochnost55 (approved):

gargshantnu (approved):

vipinsonu1 (approved):

PR #878, opened 2022-06-02
[TD-2686] Chat widget should work like Zendesk

pranavp99:
-Chat Widget Should work like Zendesk Chat where if we end chat in zendesk zopim chat and click restart conversation on chat widget a new conversation should just like how it is on zendesk

### PR Checklist

### How was the code tested?

-integrate zendesk, run chat widget locally and check if messages are coming on zopim chat or not after which end conversation from zendesk. try restarting conversation to check if a new conversation is starting or not

-

-

### DIFF
--- a/webplugin/js/app/kommunicate-ui.js
+++ b/webplugin/js/app/kommunicate-ui.js
@@ -1178,6 +1178,45 @@ KommunicateUI = {
                     'vis'
                 );   
             }
+            var restartConversation = document.getElementById(
+                'mck-restart-conversation'
+            );
+            restartConversation.addEventListener('click', function () {
+                Kommunicate.startConversation(); 
+            })
+            document.getElementById('mck-submit-comment').disabled = false;
+            kommunicateCommons.modifyClassList(
+                { class: ['mck-rating-box'] },
+                '',
+                'selected'
+            );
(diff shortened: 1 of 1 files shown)


webplugin/js/app/kommunicate-ui.js
```diff
@@ -1178,6 +1178,45 @@ KommunicateUI = {
...
+                },
+                'km-mid-conv-csat'
+            );
+            kommunicateCommons.modifyClassList(
+                {
+                    id: ['csat-1'],
+                },
+                'vis',
+                'n-vis'
+            );
```
gargshantnu, line 1219: do we want to do this change for all the conversations? both normal and zendesk chat one?
@pranavp99

gargshantnu: if no, lets try to push this code to zendesk chat related code file only.

pranavp99: no we are only doing it for zendesk, there's a checker if(convResolvedFromZendesk)

gargshantnu (approved):

mukeshthawani (approved):

Prochnost55 (approved):

PR #902, opened 2022-07-06
secure flag cookies

sidgogyan:
-add secure flag cookies

### PR Checklist

### How was the code tested?

-tested on locally

-

-

### DIFF
--- a/webplugin/js/app/km-utils.js
+++ b/webplugin/js/app/km-utils.js
@@ -257,7 +257,11 @@ KommunicateUtils = {
                 : cookiePrefix + cookie.name;
         var value = cookie.value;
         var path = '/';
-        var secure =true;
+        var secure =
+            typeof cookie.secure == 'undefined'
+                ? this.isHttpsEnabledConnection()
+                : cookie.secure;
+                
         var cookieExpiry = new Date('2038-01-19 04:14:07').toUTCString();
         var isChrome =
             navigator.userAgent.indexOf('Chrome') != -1 &&
@@ -283,7 +287,7 @@ KommunicateUtils = {
             path +
             (secure ? ';secure' : '') +
             (domain ? ';domain=' + domain : '') +
-            (';SameSite=Strict');
+            (';samesite=Strict');
     },
     getCookiePrefix: function () {
         var appOptions =
--- a/webplugin/js/app/km-utils.js
+++ b/webplugin/js/app/km-utils.js
@@ -261,7 +261,7 @@ KommunicateUtils = {
             typeof cookie.secure == 'undefined'
                 ? this.isHttpsEnabledConnection()
                 : cookie.secure;
-                
+
         var cookieExpiry = new Date('2038-01-19 04:14:07').toUTCString();
         var isChrome =
             navigator.userAgent.indexOf('Chrome') != -1 &&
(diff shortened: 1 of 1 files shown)


webplugin/js/app/km-utils.js
```diff
@@ -257,10 +257,7 @@ KommunicateUtils = {
                 : cookiePrefix + cookie.name;
         var value = cookie.value;
         var path = '/';
-        var secure =
-            typeof cookie.secure == 'undefined'
-                ? this.isHttpsEnabledConnection()
-                : cookie.secure;
+        var secure =true;
```
devashishmamgain, line 260: is there any issue with the customer's site itself is in HTTP, will everything work fine for them?

webplugin/js/app/km-utils.js
```diff
@@ -286,7 +283,7 @@ KommunicateUtils = {
             path +
             (secure ? ';secure' : '') +
             (domain ? ';domain=' + domain : '') +
-            (isChrome ? ';samesite=lax' : '');
+            (';SameSite=Strict');
```
Prochnost55, line 286: https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie
Screenshot 2022-07-07 at 12 42 49 PM

should it be SameSite or samesite?

Prochnost55 (changes requested):

webplugin/js/app/km-utils.js
sidgogyan: done

webplugin/js/app/km-utils.js
sidgogyan: i have changed code to their previous state

Prochnost55 (approved):

webplugin/js/app/km-utils.js
sidgogyan: this will add attribute secure over HTTPS and this samsite attribute of cookies will be strict

devashishmamgain:
@sidgogyan @Prochnost55 
could you please add details on why strict and not None?

please check this
Screenshot 2022-07-08 at 10 56 49 AM


https://web.dev/samesite-cookies-explained/

sidgogyan:
> @sidgogyan @Prochnost55 could you please add details on why strict and not None?
> 
> please check this Screenshot 2022-07-08 at 10 56 49 AM
> 
> https://web.dev/samesite-cookies-explained/

Sir, as per the MDN docs

[Strict](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#strict)
Cookies will only be sent in a first-party context and not be sent along with requests initiated by third party websites.

[None](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#none)
Cookies will be sent in all contexts, i.e. in responses to both first-party and cross-site requests. If SameSite=None is set, the cookie [Secure](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#secure) attribute must also be set (or the cookie will be blocked).

I thought it is not a good idea to share these cookies with third parties as these cookies are used at kommunicate side only. Therefore using strict.
Screenshot 2022-07-08 at 11 19 25 AM

devashishmamgain (approved):

PR #904, opened 2022-07-07
Removing Unnecessary anchor

Sanjeevsinghnegi:
- Removing Unnecessary anchor so that they will not show as a backlink on customers' website

### PR Checklist

### How was the code tested?

- Local Env


devashishmamgain (approved):

Prochnost55 (commented):
Should we remove the commented code if not required?

PR #915, opened 2022-07-21
[TD-2768]Add phone number in zopim(1)

pranavp99:
-Phone number details should be passed to zendesk zopim.

### PR Checklist

### How was the code tested?

-check if entered phone number is similar to the phone contact in zopim user.

-

-

### DIFF
--- a/webplugin/js/app/zendesk-chat-service.js
+++ b/webplugin/js/app/zendesk-chat-service.js
@@ -87,7 +87,7 @@ function ZendeskChatService() {
         }
     };
     _this.updateNumberInZopim = function() {
-        if(zChat.getVisitorInfo().phone != phoneNumber){
+        if(zChat.getVisitorInfo().phone != phoneNumber && phoneNumber){
             zChat.setVisitorInfo({ phone: phoneNumber }, function(err) {
                 if (!err) {
                     console.log(zChat.getVisitorInfo());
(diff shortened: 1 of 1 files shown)


webplugin/js/app/zendesk-chat-service.js
```diff
@@ -83,6 +86,15 @@ function ZendeskChatService() {
             });
         }
     };
+    _this.updateNumberInZopim = function() {
+        if(zChat.getVisitorInfo().phone != phoneNumber){
+            zChat.setVisitorInfo({ phone: phoneNumber }, function(err) {
```
devashishmamgain, line 91: this will get called even if the phone number is not provided?

devashishmamgain (changes requested):

webplugin/js/app/zendesk-chat-service.js
devashishmamgain: so now we will keep calling Zendesk APIs getVisitorInfo even if we have nothing to update

devashishmamgain (changes requested):

webplugin/js/app/zendesk-chat-service.js
pranavp99: will place an if condition above this to check if phoneNumber exists or not

pranavp99: only then it'll call getVisitorInfo

devashishmamgain: read this https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND

devashishmamgain (approved):

Prochnost55 (approved):

gargshantnu (approved):

PR #928, opened 2022-08-16
FW-4015 Media compression 25MB is the limit when being uploaded.

Prochnost55:
- instead of media compression we are providing a way which can be used to upload the videos of size > 25 mb.

### PR Checklist

### How was the code tested?

- need to add this in kommunicate settings
- attachmentHandler: function(file){}
- the function will be a custom function

- you will need a setup for testing Please connect with me for that.

-

-

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -14375,7 +14375,7 @@ var userOverride = {
 
             _this.processBeforeUpload = function (event) {
                 var fileToUpload = $applozic(this)[0].files[0];
-                if (fileToUpload.size >= 25000000) {
+                if (fileToUpload.size >= KommunicateConstants.MAX_UPLOAD_SIZE) {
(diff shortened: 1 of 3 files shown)


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -14374,6 +14373,22 @@ var userOverride = {
                 }
             };
 
+            _this.processBeforeUpload = function (event) {
+                var fileToUpload = $applozic(this)[0].files[0];
+                if (fileToUpload.size >= 25000000) {
```
gargshantnu, line 14378: magic number to constant

gargshantnu (approved):

Sanjeevsinghnegi (dismissed):

webplugin/js/app/mck-sidebox-1.0.js
Prochnost55: moved to constant files

Sanjeevsinghnegi (approved):

PR #952, opened 2022-09-22
Fix the Syntaxerror when user use pre lead collection.

anujpanwar01:
- Fix the Synaxerror when user use pre lead collection.

### PR Checklist

### How was the code tested?

- 

-

-
### Video

https://user-images.githubusercontent.com/94929705/191758866-d50baf6d-e2d8-46a7-9f7b-a3ed3c03e07d.mov

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -3142,7 +3142,7 @@ var userOverride = {
                             );
                             kmChatInput.setAttribute(
                                 'oninvalid',
-                                'setCustomValidity(' + MCK_LABELS['lead.collection'].errorEmail + ')'
+                                `setCustomValidity('${ MCK_LABELS['lead.collection'].errorEmail}')`
                             );
                             kmChatInput.setAttribute(
                                 'oninput',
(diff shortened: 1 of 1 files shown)


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -3142,7 +3142,7 @@ var userOverride = {
                             );
                             kmChatInput.setAttribute(
                                 'oninvalid',
-                                'setCustomValidity(' + MCK_LABELS['lead.collection'].errorEmail + ')'
+                                `setCustomValidity('${ MCK_LABELS['lead.collection'].errorEmail}')`
```
Prochnost55, line 3145: avoid using template strings in web-sdk
these are not supported in IE

Prochnost55 (changes requested):

webplugin/js/app/mck-sidebox-1.0.js
anujpanwar01: @Prochnost55 Done sir.

Sanjeevsinghnegi (approved):

Prochnost55 (approved):

PR #954, opened 2022-09-23
FW-4293 - Trust factor bug

Prochnost55:
-

### PR Checklist

### How was the code tested?

-

-

-


Sanjeevsinghnegi:
@Prochnost55 this is the fix of this right?
Screenshot_2022-09-20-19-43-12-55_40deb401b9ffe8e1df2f1cc5ba480b12

Sanjeevsinghnegi (approved):

PR #972, opened 2022-10-31
[TD-2651] Fixed bug (Should not create multiple threads if Zopim chat integration)

raj5036:
- Stop creating multiple threads for Zopim chat integration

### PR Checklist

### Testcases this change was tested against
- Firstly verified whether the main bug is coming anymore or not.
- Checked if new conversation is getting created when user restarts the conversation after the chat gets resolved from Zendesk. Its working. 
- Verified if new conversation is getting created after widget reload. Its working.
- Verified that user should not be able to continue with the last conversation after the chat gets resolved or reloaded.

### DIFF
--- a/webplugin/js/app/zendesk-chat-service.js
+++ b/webplugin/js/app/zendesk-chat-service.js
@@ -322,9 +322,11 @@ var onTabClickedHandlerForZendeskConversations = function (event) {
         var assigneeInfo = currentGroupData && currentGroupData.users && Object.values(currentGroupData.users).find(function (member) {
             return member.userId == currentGroupData.metadata.CONVERSATION_ASSIGNEE
         })
+        if (document.getElementById('mck-tab-title').innerHTML == "Conversations" && assigneeInfo.role != KommunicateConstants.GROUP_ROLE.MODERATOR_OR_BOT) {
+            kommunicate.startConversation();
+        }
         if (!newConversationCreated && assigneeInfo.role != KommunicateConstants.GROUP_ROLE.MODERATOR_OR_BOT) {
             newConversationCreated = true;
-            Kommunicate.startConversation();
         }
     }
 };
--- a/webplugin/js/app/zendesk-chat-service.js
+++ b/webplugin/js/app/zendesk-chat-service.js
@@ -324,7 +324,6 @@ var onTabClickedHandlerForZendeskConversations = function (event) {
         })
         if (!newConversationCreated && assigneeInfo.role != KommunicateConstants.GROUP_ROLE.MODERATOR_OR_BOT) {
             newConversationCreated = true;
-            Kommunicate.startConversation();
         }
     }
 };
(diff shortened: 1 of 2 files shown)


webplugin/js/app/zendesk-chat-service.js
```diff
@@ -324,7 +324,6 @@ var onTabClickedHandlerForZendeskConversations = function (event) {
         })
         if (!newConversationCreated && assigneeInfo.role != KommunicateConstants.GROUP_ROLE.MODERATOR_OR_BOT) {
             newConversationCreated = true;
-            Kommunicate.startConversation();
```
gargshantnu, line 327: i think we need this code, its because in zendesk we cant continue last conversation.
the bug you are looking to fix is not this one.
cross check once if zendesk is working fine with this or not

gargshantnu (changes requested):

webplugin/js/app/zendesk-chat-service.js
raj5036: The bug I was trying to fix was coming because of this line only. The bug is fixed after the removal of this line.

And when are we not suppose continue with the last conversation? After chat-widget reload?

gargshantnu: we cant continue last conversation in zendesk. its how their sdk was developed. hence this code.

after this change, are you able to continue last conversation in both direction ? zendesk - kommunicate

raj5036: No I am not being able to continue with last conversation in both direction. 

I am trying after chat resolution or/and chat-widget reload. In both cases last convo is not coming on widget.

gargshantnu (dismissed):

Prochnost55 (approved):

webplugin/js/app/zendesk-chat-service.js
```diff
@@ -322,9 +322,11 @@ var onTabClickedHandlerForZendeskConversations = function (event) {
         var assigneeInfo = currentGroupData && currentGroupData.users && Object.values(currentGroupData.users).find(function (member) {
             return member.userId == currentGroupData.metadata.CONVERSATION_ASSIGNEE
         })
+        if (document.getElementById('mck-tab-title').innerHTML == "Conversations" && assigneeInfo.role != KommunicateConstants.GROUP_ROLE.MODERATOR_OR_BOT) {
```
devashishmamgain, line 325: in case if the customer is using some other language, will this == "Conversations" still work?

devashishmamgain (changes requested):

webplugin/js/app/zendesk-chat-service.js
raj5036: But do we use translations on widget (like we do on Dashboard)? How can it get translated?

devashishmamgain: check this https://docs.kommunicate.io/docs/web-localization#chat-localization-setup

raj5036: Updated. 
Picking the value dynamically now. instead of hardcoding

devashishmamgain (dismissed):

PR #985, opened 2022-11-22
added readme file

surajkukkipady18:
-

### PR Checklist

### How was the code tested?

-

-

-

### DIFF
--- a/playwright_test/README.md
+++ b/playwright_test/README.md
@@ -1,0 +1,61 @@
+
+This Project is created in [Playwright](https://playwright.dev/)
+
+## Contents:
+
+- [Install PlayWright](#installing-playwright)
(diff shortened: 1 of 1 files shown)


playwright_test/README.md
```diff
@@ -0,0 +1,61 @@
...
+## Installing Playwright
+Get started by installing Playwright using npm.
+
+```js
+npm init playwright@latest
+```
+
+Run the install command and select the following to get started:
+
+- Choose between TypeScript or JavaScript (default is TypeScript)
```
devashishmamgain, line 19: which one are we using? mention that

devashishmamgain (changes requested):

devashishmamgain (approved):

PR #996, opened 2022-12-16
Fixed Bug for CreatedAt giving Undefined

Sanjeevsinghnegi:
- Fixed Bug for CreatedAt giving Undefined

### PR Checklist

### How was the code tested?

-

### ScreenShot
- Bug
Screenshot 2022-12-16 at 9 29 50 PM

### In case you fixed a bug then please describe the root cause of it. 
- there is no data check and we are getting undefined 

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -6390,7 +6390,7 @@ var userOverride = {
                     success: function (data) {
                         var isMessages = true;
                         //Display/hide lead(email) collection template
-                        CURRENT_GROUP_DATA.createdAt = data.groupFeeds[0].createdAtTime;
+                        CURRENT_GROUP_DATA.createdAt = data && data.groupFeeds[0].createdAtTime;
                         CURRENT_GROUP_DATA.isgroup = params.isGroup;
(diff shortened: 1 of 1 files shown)


devashishmamgain (commented):
not sure if it will fix the issue, as far as i understand error is coming for groupFeeds to be empty and not data as undefined.
is it tested by reproducing?

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -6390,7 +6390,7 @@ var userOverride = {
                     success: function (data) {
                         var isMessages = true;
                         //Display/hide lead(email) collection template
-                        CURRENT_GROUP_DATA.createdAt = data.groupFeeds[0].createdAtTime;
+                        CURRENT_GROUP_DATA.createdAt = data && data.groupFeeds[0].createdAtTime;
```
devashishmamgain, line 6393: data &&
                            data.groupFeeds[0] &&

same as the code below it at line no. 6395

devashishmamgain (changes requested):

Sanjeevsinghnegi:
Closing this as @Prochnost55 also fixed the same bug

PR #997, opened 2022-12-17
Fix the bug where the widget stuck in loading state on click of back button

Prochnost55:
- The widget was getting stuck in loading state when the back button was getting clicked
- this was happening in the normal flow

### PR Checklist

### How was the code tested?

- Created a multiple conversations in the normal flow
- created multiple conversations in zopim flow.

-

- a check was missing for the groupFeeds array


devashishmamgain (approved):

Sanjeevsinghnegi (approved):

Sanjeevsinghnegi:
@Prochnost55 multiple same checks  are in the code later we can add only one check

gargshantnu (approved):

PR #1004, opened 2022-12-28
Automate Common checks in PR [Development] [FWH-40]

Sanjeevsinghnegi:
- Automate Common checks in PR

### PR Checklist

### How was the code tested?

-

- danger.js

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -12,6 +12,7 @@ const PR_REVIEWS_CHECKS = Object.freeze({
         WIP: `PR is classed as Work in Progress`,
         EXCEED_THRESHOLD: `:exclamation: Your PR has over ${BIG_PR_THRESHOLD} lines of code, please consider breaking it down so we can effectively review it. :thumbsup:`,
         MERGE_CONFLICTS: `Please merge development branch`,
+        ONLY_DEV_BRANCH: `:exclamation: Please re-submit this PR to development branch, we may have already fixed your issue. 🙏`,
         MISSING_DES: `:exclamation: Please add a description to your PR to make it easier to review`,
         MERGE_COMMITS: `Please rebase to get rid of the merge commits in this PR 🙏`,
         FILE_ISSUE: `Please file the issue for development branch`,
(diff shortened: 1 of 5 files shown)


Sanjeevsinghnegi:
@abhishek7214 there is no package-lock.json file in this, also it is under .gitignore should i commit the same?

dangerfile.js
```diff
@@ -0,0 +1,71 @@
...
+        PACKAGE_JSON_ERR: `Changes were made to package.json, but not to package-lock.json - <i>'Perhaps you need to run \`npm install\`?'`
+    },
+    SUCCESS: {
+        LINES_REMOVED : `:tada: The PR removed ${danger.github.pr.deletions} lines. :clap:`,
+        NEW_FILES: `New Files in this PR: \n - `
+    }
+});
+
+//Accept PR only for development branch
+if(danger.github.pr.head.ref == "master") return fail(PR_REVIEWS_CHECKS.WARNS.ONLY_DEV_BRANCH);
```
prochnost04, line 30: At times we need to file PR for master branch

dangerfile.js
```diff
@@ -0,0 +1,71 @@
...
+// check PR is changes Threshold
+const removedLines = danger.github.pr.deletions,
+    totalChanges = danger.github.pr.additions + removedLines > PR_REVIEWS_CHECKS.BIG_PR_THRESHOLD;
+totalChanges && warn(PR_REVIEWS_CHECKS.WARNS.EXCEED_THRESHOLD);
+
+// check PR delete lines status
+if(removedLines > PR_REVIEWS_CHECKS.REMOVED_LINE_THRESHOLD) message(PR_REVIEWS_CHECKS.SUCCESS.LINES_REMOVED);
+
+// check if PR has some merge conflicts
+const mergeCommits = danger.github.commits.filter(({ commit }) => commit.message.includes(`Merge branch 'development'`));
```
prochnost04, line 52: can this cause trouble when we make a release branch?

package.json
```diff
@@ -38,6 +40,9 @@
     "node-fetch": "2.6.7"
   },
   "devDependencies": {
+    "danger": "^11.2.0",
```
prochnost04, line 43: we can remove ^

dangerfile.js
Sanjeevsinghnegi: in that case, it will only check this condition and return

if you want i will remove return from here

package.json
Sanjeevsinghnegi: ok

dangerfile.js
Sanjeevsinghnegi: in that case it will show only warning and then checks for other conditions also

prochnost04: Ok so 
1. will it not fail the PR creation? 
2. The action bot will not say "checks failed"?

We can add warn and remove return

Sanjeevsinghnegi: 1. will it not fail the PR creation? NO
2. The action bot will not say "checks failed"? adding warn

dangerfile.js
Sanjeevsinghnegi: as i checked no

prochnost04 (approved):

panwaranuj01 (approved):

PR #1007, opened 2023-01-05
Add query parameters for iframe [FFH-49]

sonujangra27:
- Add query parameters for iframe

### PR Checklist

### How was the code tested?

- Pre Chat Lead on from dashboard - Chat widget opens with Pre Chat Lead Form
 If Disabled - Conversation starts

-

-

### DIFF
--- a/webplugin/template/chat.html
+++ b/webplugin/template/chat.html
@@ -48,18 +48,17 @@
                                 iframe.classList.add("change-kommunicate-iframe-height");
                             });
                         });
-                        //to show the pre chat lead collection Form if enabled 
+                        //to show the pre chat lead collection Form if enabled
                         var preChatLauncher = window.KommunicateGlobal.window.document.getElementById("km-anonymous-chat-launcher")
                         if (window.KommunicateGlobal && preChatLauncher) {
                             var intervalForPreChatLead = setTimeout(function () {
                                 preChatLauncher.click()
                                 window.KommunicateGlobal.window.document.getElementById("km-modal-close").style.display = "none"
                             }, 500);
-                        }
-                        window.KommunicateGlobal && preChatLauncher &&
                             preChatLauncher.addEventListener("click", function () {
                                 { clearInterval(intervalForPreChatLead) }
                             })
+                        }
                         closeButton.parentElement.style.display = "none";
                         Kommunicate.launchConversation();
                         // KommunicateGlobal.document.getElementById('mck-sidebox-launcher').click();
--- a/webplugin/template/chat.html
+++ b/webplugin/template/chat.html
@@ -48,9 +48,21 @@
                                 iframe.classList.add("change-kommunicate-iframe-height");
                             });
                         });
+                        //to show the pre chat lead collection Form if enabled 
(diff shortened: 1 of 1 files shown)


webplugin/template/chat.html
```diff
@@ -13,15 +13,25 @@
...
 
         if (!appId) {
             document.getElementById('error-msg').innerText = "no app id"
         } else {
             (function (d, m) {
+                if(botIds && assignee ){
+                var defaultSettings = {
+                "defaultBotIds": [botIds], // Optional. Replace <BOT_ID> with your bot ID which you can find in bot section of dashboard. Include this parameter if you want to assign the conversation to a bot.
+                "defaultAssignee": assignee, // Replace <ASSIGNEE_ID> with your bot/agent ID  whom you're assigning the conversation.
+                "skipRouting": true
```
prochnost04, line 28: ideally this should be optional
default value being true

sonujangra27: My understanding was this if we make it true then only it will skip the Dashboard routing rules then and use the assignees defined here - 
if we make it false or eradicate - will the setting still apply?

prochnost04: > if we make it false or eradicate - will the setting still apply?
Please check by keeping it false and update accordingly

sonujangra27: if it is false - it routes to default, we will need to keep it as True @prochnost04

prochnost04: okay

prochnost04 (approved):

webplugin/template/chat.html
```diff
@@ -38,9 +48,21 @@
                                 iframe.classList.add("change-kommunicate-iframe-height");
                             });
                         });
+                        //to show the pre chat lead collection Form if enabled 
+                        var preChatLauncher = window.KommunicateGlobal.window.document.getElementById("km-anonymous-chat-launcher")
+                        if (window.KommunicateGlobal && preChatLauncher) {
+                            var intervalForPreChatLead = setTimeout(function () {
+                                preChatLauncher.click()
+                                window.KommunicateGlobal.window.document.getElementById("km-modal-close").style.display = "none"
```
prochnost04, line 56: can you add the clearInterval here itself?

webplugin/template/chat.html
```diff
@@ -38,9 +48,21 @@
...
+                            }, 500);
+                        }
+                        window.KommunicateGlobal && preChatLauncher &&
+                            preChatLauncher.addEventListener("click", function () {
+                                { clearInterval(intervalForPreChatLead) }
+                            })
                         closeButton.parentElement.style.display = "none";
                         Kommunicate.launchConversation();
-                        KommunicateGlobal.document.getElementById('mck-sidebox-launcher').click();
+                        // KommunicateGlobal.document.getElementById('mck-sidebox-launcher').click();
```
prochnost04, line 65: will this work for normal use case
ie were prechat lead form is not there?

webplugin/template/chat.html
sonujangra27: done

webplugin/template/chat.html
sonujangra27: yes, I checked!

prochnost04 (approved):

PR #1019, opened 2023-01-20
Added test for List Template and updated snapshot match method

archit2602:
- Testing List Template Rich message UI
- Updated all the toMatchSnapshot() function and reduced threshold for better match

### PR Checklist

### How was the code tested?

- navigate to `cd playwright_test/`
- run `npx playwright test`

-

-


playwright_test/tests/richMessage_Combination_buttons.spec.js
```diff
@@ -45,7 +45,11 @@ let page;
     const iframe = page.frameLocator(widgetLocators.kmIframe)
 
   // The UI of rich message check using snapshot
-    expect(await page.screenshot()).toMatchSnapshot('Button_Combination.png',{threshold: 1});
+    const screenshot = await page.screenshot();
+    expect(screenshot).toMatchSnapshot({
+      threshold: 0.98,
+      name: 'Button_Combination.png'
```
devashishmamgain, line 51: move threshold to a constant so that its valued can be tweaked some a common place

devashishmamgain (changes requested):

playwright_test/tests/richMessage_Combination_buttons.spec.js
archit2602: What do you mean by that ??

devashishmamgain: lets say if we want to change threshold to 0.8 in future, do we need to go through each and every test code and then keep modifying the value?
or there can be a constant defined someone in the common code, modifying that changes it every test case  
https://www.w3schools.com/js/js_const.asp

archit2602: oohhh that constant I was thinking about constant value 
ya ill do that

archit2602: added

playwright_test/locaterPackage/kmLocators.js
```diff
@@ -40,3 +40,7 @@ export const richMessagesLocators = {
     kmSubmitButtonResponse: '//div[text()="Button Response"]',
     kmImageResponse: '//div[@class="km-image-template"]'
 };
+
+export const commonValue = {
```
devashishmamgain, line 44: make it COMMON_VALUES

archit2602: can't start with capital letters ill make that common_Values

archit2602: done

devashishmamgain: why not?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const

archit2602: I mean it is not recommended to start with a capital letter if I convert this into capital then I have to convert all the names to capital.
If you say ill do that I don't have any issue

devashishmamgain: "not recommended" 
>> where did you read this? as far as i understand, const are usually named in all CAPS so that it is easy to understand for other developers that this is a constant, and not a variable

archit2602: I have gone through a lot of GitHub codes where people are using this method only but it fil ill do the change 😊

devashishmamgain (approved):

vipinkom1 (approved):

PR #1020, opened 2023-01-25
Added Test for card, HTML and Video rich messages

archit2602:
- Testing HTML, Card, and video Rich message

### PR Checklist

### How was the code tested?

- navigate to `cd playwright_test/`
- run `npx playwright test`

-

-


playwright_test/tests/richMessage_Card_Carousel.spec.js
```diff
@@ -0,0 +1,56 @@
...
+import { test, expect } from '@playwright/test';
+import { WIDGET_LOCATORS, LOCATORS, COMMON_VALUES } from '../locaterPackage/kmLocators.js';
+import { URL , APP_ID } from '../utils/kmSecret';
+import { SCRIPT } from '../utils/kmScript';
+
+let page;
+
+// Launching widget
+  test.beforeAll(async ({browser}) => {
+    page = await browser.newPage();
```
devashishmamgain, line 16: is this same launch code be which is present in other test cases as well? If yes, move it to a common code file and call it from all places

devashishmamgain (changes requested):

playwright_test/tests/richMessage_Card_Carousel.spec.js
archit2602: ya this is the same but when we add more test cases in the same file we need this to be in the same file because we can shift some of the test code to beforeAll()

devashishmamgain: may be move common code to a separate file, and then call that function from everywhere

archit2602: if we can take this in last because have to do a lot of trials to separate beforeAll() and afterAll() with this getting lots of errors due to the playwright version and other stuff I will take this in last as code formatting ?? if you are ok with this

devashishmamgain: ok

devashishmamgain (approved):

vipinkom1 (approved):

PR #1031, opened 2023-02-06
Iframe Width Issue [FFH-115]

sonujangra27:
- Fix  to make chat widget full screen  when using through Iframe- currently, after 700px or larger it was breaking

### PR Checklist

### How was the code tested?

- Without pre chat lead -
 width give 100%
 
 with pre chat lead enabled
 test for height and width 100%

-

-

### DIFF
--- a/webplugin/template/chat.html
+++ b/webplugin/template/chat.html
@@ -53,15 +53,20 @@
                         if (window.KommunicateGlobal && preChatLauncher) {
                             var intervalForPreChatLead = setTimeout(function () {
                                 preChatLauncher.click()
+                                iframe.classList.add("change-kommunicate-iframe-height");
+                                var preChatLeadScreenId = window.KommunicateGlobal.window.document.getElementById("km-chat-login-modal")
+                                var cssForPreChatLeadScreen = "#km-chat-login-modal{height:100%!important;width:100%!important;right:0!important;bottom:0!important;max-height: 100%!important;}"
+                                preChatLeadScreenId && Kommunicate.customizeWidgetCss(cssForPreChatLeadScreen);
                                 window.KommunicateGlobal.window.document.getElementById("km-modal-close").style.display = "none"
-                            }, 500);
+                            }, 100);
                             preChatLauncher.addEventListener("click", function () {
                                 { clearInterval(intervalForPreChatLead) }
                             })
                         }
                         closeButton.parentElement.style.display = "none";
                         Kommunicate.launchConversation();
-                        // KommunicateGlobal.document.getElementById('mck-sidebox-launcher').click();
+                        var launcherIcon = KommunicateGlobal.document.getElementById('mck-sidebox-launcher')
+                        launcherIcon && launcherIcon.click();
                     }
                 };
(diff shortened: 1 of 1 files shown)


webplugin/template/chat.html
```diff
@@ -53,15 +53,20 @@
...
-                            }, 500);
+                            }, 100);
                             preChatLauncher.addEventListener("click", function () {
                                 { clearInterval(intervalForPreChatLead) }
                             })
                         }
                         closeButton.parentElement.style.display = "none";
                         Kommunicate.launchConversation();
-                        // KommunicateGlobal.document.getElementById('mck-sidebox-launcher').click();
+                        var launcherIcon = KommunicateGlobal.document.getElementById('mck-sidebox-launcher')
```
Sanjeevsinghnegi, line 68: in line no 58 you are checking for window.KommunicateGlobal bot in 68 not why?

Sanjeevsinghnegi (changes requested):

webplugin/template/chat.html
```diff
@@ -53,15 +53,20 @@
                         if (window.KommunicateGlobal && preChatLauncher) {
                             var intervalForPreChatLead = setTimeout(function () {
                                 preChatLauncher.click()
+                                iframe.classList.add("change-kommunicate-iframe-height");
+                                var preChatLeadScreenId = window.KommunicateGlobal.window.document.getElementById("km-chat-login-modal")
+                                var cssForPreChatLeadScreen = "#km-chat-login-modal{height:100%!important;width:100%!important;right:0!important;bottom:0!important;max-height: 100%!important;}"
+                                preChatLeadScreenId && Kommunicate.customizeWidgetCss(cssForPreChatLeadScreen);
                                 window.KommunicateGlobal.window.document.getElementById("km-modal-close").style.display = "none"
-                            }, 500);
+                            }, 100);
```
Sanjeevsinghnegi, line 61: what is the purpose of 100 here

sonujangra27: I am checking as soon as this item is available - if it check it on 500ms interval - I see a delay jisse icon dikh rha hai

webplugin/template/chat.html
sonujangra27: because when pre chat lead is not enabled it is not necessary - this div is available

there I am checking when the chat widget is loading

webplugin/template/chat.html
Sanjeevsinghnegi: you mean after 100ms you are checking something?

sonujangra27: yes - 
i am checking for the id so that I can show the prechatlead window

var preChatLauncher = window.KommunicateGlobal.window.document.getElementById("km-anonymous-chat-launcher")

Sanjeevsinghnegi (approved):

webplugin/template/chat.html
devashishmamgain: what is the guarantee that it will be done within 100 ms? what will happen if this assumption goes wrong for any reason?

sonujangra27: it will check every 100ms second

devashishmamgain (approved):

PR #1032, opened 2023-02-07
Fwh-67 dev Audio Message contentType: 8 handling | Web SDK

pranav4-4:
- when we send audio from mobile it should be handled in web sdk

### PR Checklist

### How was the code tested?

- I have tested it from https://test.prochnost.tk/widget-test

-

-

### DIFF
--- a/webplugin/js/app/kommunicate.js
+++ b/webplugin/js/app/kommunicate.js
@@ -199,14 +199,7 @@ $applozic.extend(true, Kommunicate, {
     },
     openConversation: function (groupId) {
         kommunicateCommons.setWidgetStateOpen(true);
-        if (groupId) {
-            window.$applozic.fn.applozic('loadGroupTab', groupId);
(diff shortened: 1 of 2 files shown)


webplugin/js/app/kommunicate.js
```diff
@@ -199,7 +199,14 @@ $applozic.extend(true, Kommunicate, {
...
         kommunicateCommons.setWidgetStateOpen(true);
-        window.$applozic.fn.applozic('loadGroupTab', groupId);
+        if (groupId) {
+            window.$applozic.fn.applozic('loadGroupTab', groupId);
+        } else if (groupDetail && groupDetail.clientGroupId) {
+            window.$applozic.fn.applozic(
+                'loadGroupTabByClientGroupId',
+                groupDetail
+            );
+        }
```
prochnost04, line 209: @pranav4-4 we can revert this change

prochnost04 (approved):

PR #1034, opened 2023-02-08
 If the encrypted user uses the pre chat lead then API is not encrypted in type 3 and in type 5 widget not opening

panwaranuj01:
- If the encrypted user uses the pre-chat lead then API is not encrypted in type 3 and in type 5 widget not opening
- this is because we don't ` enableEncryption` property in the payload

### PR Checklist

### How was the code tested?

- Test for pre chat encrypted account [type 3,5]
- without prechat lead in encrypted account
- Test without encrypted account also

-

-

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -4825,6 +4825,7 @@ var userOverride = {
                         googleApiKey: MCK_GOOGLE_API_KEY,
                         chatNotificationMailSent: true,
                         authenticationTypeId: MCK_AUTHENTICATION_TYPE_ID,
+                        enableEncryption: true
                     };
                     if (email) {
                         options.email = email.toLowerCase();


Sanjeevsinghnegi (approved):

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -4825,6 +4825,7 @@ var userOverride = {
                         googleApiKey: MCK_GOOGLE_API_KEY,
                         chatNotificationMailSent: true,
                         authenticationTypeId: MCK_AUTHENTICATION_TYPE_ID,
+                        enableEncryption: true
```
Prochnost55, line 4828: what will happen if the account is not encrypted?
is that working fine?

prochnost04 (commented):
What will happen if the account is not ecrypted
Is it working fine for that case?

panwaranuj01:
> What will happen if the account is not encrypted
> Is it working fine for that case?

Working fine

prochnost04 (approved):

PR #1059, opened 2023-04-13
TTS for IOS mobile browser is not working [TDH-255]

Sanjeevsinghnegi:
- Code Improvements TTS
- Fix IOS devices TTS not working

### PR Checklist

### How was the code tested?

- Checked on Mac (safari, chrome, firefox, edge)
- Checked on Windows (chrome, firefox, edge)
- Checked on Android Web (chrome, firefox, edge)
- Checked on IOS Web (Safari)

### In case you fixed a bug then please describe the root cause of it. 
- IOS devices did not run speech API automatically, so manually triggering the fake empty voice for IOS

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -890,12 +890,29 @@ var userOverride = {
             // the browser call getVoices is async
             // so we are updating the array whenever they're available
             if (VOICE_OUTPUT_ENABLED && "speechSynthesis" in window) {
+                var isIosDevice = /iPhone|iPad|iPod/i.test(navigator.userAgent) || false;
+
                 AVAILABLE_VOICES_FOR_TTS = speechSynthesis.getVoices();
                 if (speechSynthesis.onvoiceschanged !== undefined) {
                     speechSynthesis.onvoiceschanged = function () {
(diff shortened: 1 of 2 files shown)


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -890,12 +890,29 @@ var userOverride = {
...
+                    /** 
+                        it is only for IOS devices so using newer syntax
+                        IOS devices would not let the speech API run programmatically unless we have triggered manually one time under the user's interaction.
+                     */
+
+                    const simulateFakeVoice = () => {
+                        const fakeInstance = new SpeechSynthesisUtterance('');
+                        speechSynthesis.speak(fakeInstance);
+                    };
+                    document.addEventListener('click', simulateFakeVoice, {
```
prochnost04, line 911: unless a user clicks the, speech will not work right?

Sanjeevsinghnegi: yes, so I am considering the click-on Iframe document only once, so if a user initialize the widget we are good to go with TTS.
FYI: TTS is going to start work from welcome message itself

prochnost04 (approved):

panwaranuj01 (approved):

PR #1094, opened 2023-07-26
TDH-338 Pseudonym appearing in place of Name in first chats sometimes

Aryan-Kirsali:
- When a chat is initiated for a new user for the first time, in some cases the pseudonym assigned to that user is displaying in place of his/her name in the chat.

### PR Checklist

### How was the code tested?

- By running the widget in my local host and changing the config-env.js file to use my local code for testing in Chrome browser. 

- Asynchronous network calls and functional programming

- The cause of the problem was the updateUser function. This function is an asynchronous function which uses AJAX to send a request and await its response, but also it lets the javascript code to move forward irrespective of its response. The code doesn't wait for the response.
- So what's happening is, in some cases the send message request is getting processed before the updateUser request due to which the entity extractor is extracting the pseudonym(cause the user Update is not done yet)

### Screenshot
- Below is the proper sequence in which requests should process
- Trigger Intent

Screenshot 2023-07-26 at 1 21 11 PM

- Update Name

Screenshot 2023-07-26 at 1 21 22 PM

- Send Message

Screenshot 2023-07-26 at 1 21 32 PM

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -5341,63 +5341,6 @@ var userOverride = {
                         $mck_box_form.addClass('mck-text-req');
                         return false;
                     }
-                    //If the field is a form field then validate the input, update user details before sending the message
-                    if ($mck_text_box.data('fieldType')) {
-                        //If the field has a regex validation then validate the input otherwise skip validation
-                        if($mck_text_box.data('validation')){
-                            var regexForm = $mck_text_box.data(
-                                'validation'
-                            );
-                            regexForm = new RegExp(regexForm);
-                            //If the input does not match the regex validation then show an error message for 2 seconds.
-                            if (!regexForm.test(message)) {
-                                $applozic('#mck-form-field-error-alert').html($mck_text_box.data('errorMessage'));
-                                $applozic('#mck-form-field-error-alert-box')
-                                    .removeClass('n-vis')
-                                    .addClass('vis');
-                                setTimeout(function () {
-                                    $applozic('#mck-form-field-error-alert-box')
-                                        .removeClass('vis')
-                                        .addClass('n-vis');
-                                }, 2000);
-                                return false;
-                            }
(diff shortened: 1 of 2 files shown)


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -5485,9 +5428,95 @@ var userOverride = {
...
+                                    data: { metadata: userUpdateField },
+                                    success: function () {
+                                        sendMessageWhenFieldType();
+                                    },
+                                    error: function () {
+                                        sendMessageWhenFieldType();
+                                    }
+                                });
+                            }
+                        }
```
prochnost04, line 5484: We can improve this logic.
1. you can use switch case for $mck_text_box.data('fieldType')
2. mckContactService.updateUser() is repeating. the data, which is being sent, is different in both places. We can avoid duplication.

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -5485,9 +5428,95 @@ var userOverride = {
...
+                            var fieldVal = $mck_text_box.data('field');
+                            var userUpdateField = {};
+                            userUpdateField[fieldVal] = message;
+                            if (
+                                $mck_text_box.data('fieldType') === 'EMAIL' ||
+                                $mck_text_box.data('fieldType') === 'NAME' ||
+                                $mck_text_box.data('fieldType') ===
+                                    'PHONE_NUMBER'
+                            ) {
+                                mckContactService.updateUser({
```
prochnost04, line 5464: Also, see we can avoid this update user call altogether. 
There are 2 issues
1. when bot asks for name and instead of only name, you say "my name is abhishek" then "my name is abhishek" gets updated as name of the user
2. if updateUser() call fails then we again face the same issue.
```suggestion
                                mckContactService.updateUser({
```

prochnost04 (changes requested):

panwaranuj01 (approved):

PR #1101, opened 2023-08-15
Sending events to Google Analytics4 as per gtag [TDH-331]

Sanjeevsinghnegi:
- Sending events to Google Analytics4 as per gtag

### PR Checklist

### How was the code tested?

- By adding ga4 script on local website then intialize the script on that website(the same ga4 id i added on dashboard)

- ga4 new event firing [instructions](https://developers.google.com/analytics/devguides/collection/ga4/events?client_type=gtag)

### ScreenRecording
- 

### DIFF
--- a/webplugin/js/app/km-widget-custom-events.js
+++ b/webplugin/js/app/km-widget-custom-events.js
@@ -83,7 +83,7 @@ var eventMapping = {
         data: {
             eventCategory: 'Kommunicate_widget',
             eventAction: 'km_widget_close',
-            eventLabel: 'Km Widget Chat Widget Close',
+            eventLabel: 'Chat Widget Close',
         },
         eventFunction: null,
     },
@@ -91,7 +91,7 @@ var eventMapping = {
         data: {
             eventCategory: 'Kommunicate_widget',
             eventAction: 'km_widget_rich_message_button_click',
-            eventLabel: 'Km Widget RichMessage Button Click',
+            eventLabel: 'RichMessage Button Click',
         },
         eventFunction: null,
     },
@@ -99,7 +99,7 @@ var eventMapping = {
         data: {
             eventCategory: 'Kommunicate_widget',
             eventAction: 'km_widget_faq_click',
-            eventLabel: 'Km Widget FAQ Menu',
+            eventLabel: 'FAQ Menu',
         },
         eventFunction: null,
     },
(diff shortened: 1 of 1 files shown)


devashishmamgain:
@Sanjeevsinghnegi are we checking if the customer is using GA4 or older version?

webplugin/js/app/km-widget-custom-events.js
```diff
@@ -12,24 +12,16 @@ var kmWidgetEvents = {
...
+                typeof window.top.gtag !== 'undefined'
             ) {
-                window.top.ga('create', kmWidgetEvents.gaTrackingId(), 'auto');
-                window.top.ga('send', {
-                    hitType: 'event',
-                    eventCategory: eventCategory,
-                    eventAction: eventAction,
-                    eventLabel: eventLabel,
-                    eventValue: eventValue,
+                window.top.gtag('event', eventObject.data.eventAction, {
```
prochnost04, line 21: do we need to update any 3rd party script also?

Sanjeevsinghnegi: no, basically we are sending events to the user's website ga layer from the widget. there is no direct relation between chat widget and google analytics. Hope this helps. if not let me know

Sanjeevsinghnegi:
> @Sanjeevsinghnegi are we checking if the customer is using GA4 or an older version?

@devashishmamgain 
**Starting July 1, 2023, Universal Analytics properties will stop processing data**

so here we only care about GA4, and we stop supporting UA(Universal analytics) from now as data is not showing there.
If the customer is using UA(older version) then they need to update to ga4

webplugin/js/app/km-widget-custom-events.js
```diff
@@ -80,152 +67,151 @@ var kmWidgetEvents = {
...
+            eventLabel: 'Km Widget CSAT Submit',
         },
         eventFunction: null,
     },
     onShowResolvedClick: {
         data: {
             eventCategory: 'Kommunicate_widget',
             eventAction: 'km_widget_resolve_click',
-            eventLabel: 'Show Resolve',
+            eventLabel: 'Km Widget Show Resolve',
```
devashishmamgain, line 126: from where these names came? what was the issue with earlier label?

devashishmamgain (changes requested):
are we checking if the customer is using older version or newer?

devashishmamgain (dismissed):

webplugin/js/app/km-widget-custom-events.js
Sanjeevsinghnegi: in ga4 only event action is the only field required so i am sending eventCategory, eventLabel as an additional property under that event action so i added this but it is not req to change the eventLabel so i removed the added prefix

prochnost04 (dismissed):

panwaranuj01 (approved):

prochnost04 (approved):

devashishmamgain (approved):

PR #1161, opened 2024-01-16
Added the es6 support via gulp

panwaranuj01:
- Added the es6 support via gulp
- currently we are using gulp only for build not for development

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 


Sanjeevsinghnegi:
@anujpanwar01 do we need to check this?
Changes were made to package.json, but not to package-lock.json - 'Perhaps you need to run npm install?'

panwaranuj01:
> @anujpanwar01 do we need to check this? Changes were made to package.json, but not to package-lock.json - 'Perhaps you need to run npm install?'

Because package-lock.json is in gitignore

prochnost04:
> > @anujpanwar01 do we need to check this? Changes were made to package.json, but not to package-lock.json - 'Perhaps you need to run npm install?'
> 
> Because package-lock.json is in gitignore

We should not be keeping package-lock.json in .gitignore
I recently came across multiple articles mentioning same thing.
The reason is, there would be "this is working fine on my machine" issues

Sanjeevsinghnegi (approved):

prochnost04 (approved):

PR #1162, opened 2024-01-17
update the readme file to run the websdk

panwaranuj01:
- update the readme file to run the websdk in local
- debug the production websdk in localhost

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -11,11 +11,12 @@ Step-1: Open Terminal on your machine.
 
 Step-2: Change the current working directory to Kommunicate-Web-SDK directory.
 
-Step-3: Move to the server folder in Kommunicate-Web-SDK directory.
-> $ cd server
+Step-3: Open the terminal and run this command to install the dependencies
+> $ npm install
 
-Step-4  To run the node server run the below command.
-> $ NODE_ENV=prod node app
+Step-3: To run the Kommunicate-Web-SDK in local you need to run this command in terminal
+> $ npm run dev
+
 
 Step-5: Open your web browser and enter http://localhost:3030 to run the demo.
 
--- a/server/README.md
+++ b/server/README.md
@@ -6,7 +6,7 @@
 - Clone Kommunicate-Web-SDK repository on your machine. ([How to clone a Github repository ?](https://help.github.com/en/github/creating-cloning-and-archiving-repositories/cloning-a-repository))
 
 
-## Steps to deploy Kommunicate server.
+## How to run Kommunicate-Web-SDK in localhost.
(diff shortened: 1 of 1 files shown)


prochnost04 (approved):

server/README.md
```diff
@@ -6,19 +6,40 @@
...
+> $ npm run dev
 
-Step-4  To run the node server run the below command.
-> $ NODE_ENV=prod node app
 
 Step-5: Open your web browser and enter http://localhost:3030 to run the demo.
 
+## How to debug the Production Kommunicate-Web-SDK in localhost.
+
+> Note -> These steps will be applicable for all environments like -> test, release, prod_in, prod etc. Make sure update only that environment key and hostUrl.
```
devashishmamgain, line 25: "Make sure update only that environment key and hostUrl." 
->
Please ensure that you update only the environment key and hostUrl

server/README.md
```diff
@@ -6,19 +6,40 @@
...
+
+Step-1: Open the `config-env.js` file in Kommunicate-Web-SDK.
+
+Step-2: Search for the `development:` key in `config-env.js` file an replace the `development -> _development`
```
devashishmamgain, line 29: file an replace
>> is it correct, or you mean file, and replace

devashishmamgain: couldn't understand why to replace it with prefixing with an _

server/README.md
```diff
@@ -6,19 +6,40 @@
...
-Step-4  To run the node server run the below command.
-> $ NODE_ENV=prod node app
 
 Step-5: Open your web browser and enter http://localhost:3030 to run the demo.
 
+## How to debug the Production Kommunicate-Web-SDK in localhost.
+
+> Note -> These steps will be applicable for all environments like -> test, release, prod_in, prod etc. Make sure update only that environment key and hostUrl.
+
+Step-1: Open the `config-env.js` file in Kommunicate-Web-SDK.
```
devashishmamgain, line 27: remove - from all Steps
directly write
Step 1:

devashishmamgain (changes requested):

server/README.md
panwaranuj01: ok

server/README.md
panwaranuj01: If we don't add the prefix before the development key it will always consider the first development object configuration. User configuration will not work in this case.

either we can comment out the first development object configuration or we can replace that key with some random letters.

devashishmamgain (approved):

PR #1166, opened 2024-02-02
FWH-1297 -: Widget loader not getting hidden after the response.

panwaranuj01:
- Widget loader not getting hidden after the response.

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 


webplugin/js/app/components/typing-service.js
```diff
@@ -0,0 +1,100 @@
+function TypingService(appOptions) {
```
prochnost04, line 1: it is a typing indicator service
also see if we should use class instead of a constructor function

panwaranuj01: will change this in class when es6 support pr will be merge

Sanjeevsinghnegi (approved):

prochnost04 (dismissed):

prochnost04 (approved):

PR #1175, opened 2024-03-13
TDH-765 -: Bot Generative AI- Phase 1 (Widget)

panwaranuj01:
- Bot Generative AI- Phase 1 (Widget)

### PR Checklist

### How was the code tested?

- tokenize response is supported only in custom bot for now 
- `generativeResponse` parameter should be true in  the bot.

https://github.com/Kommunicate-io/Kommunicate-Web-SDK/assets/109517510/0b7e7a23-e11e-46bc-81d5-2f436e43891c

-

-

### Screenshot
- 

### DIFF
--- a/webplugin/js/app/conversation/gen-ai-service.js
+++ b/webplugin/js/app/conversation/gen-ai-service.js
@@ -1,0 +1,43 @@
+class GenAiService {
+    constructor() {
+        this.currentElement = null;
+        this.textMsgDiv = null;
+    }
+
+    addTokenizeMsg = (...args) => {
+        const [msg, className, $textMessage] = args;
+
+        if (!this.currentElement) {
+            this.currentElement = document
+                .querySelector(`div[data-msgkey="${msg.key}"]`)
+                ?.querySelector(`.${className}`);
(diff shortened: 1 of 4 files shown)


webplugin/js/app/conversation/gen-ai-service.js
```diff
@@ -0,0 +1,43 @@
+class GenAiService {
+    constructor() {
+        this.currentElement = null;
+        this.textMsgDiv = null;
+    }
+
+    addTokenizeMsg = (...args) => {
+        const [msg, className, $textMessage] = args;
+
+        if (!this.currentElement) {
```
devashishmamgain, line 10: is it possible that the currentElement exists but for a different message key, and not for the msg.key which is passed?

panwaranuj01: no, this case will not happen because I'm resetting the currentElement and textMsgDiv values when the user sends the message.

devashishmamgain (approved):

prochnost04 (approved):

PR #1186, opened 2024-04-27
Fwh 1632 Chat Widget CSAT Changes

AashishBhatt5:
-

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 


webplugin/js/app/labels/default-labels.js
```diff
@@ -144,7 +145,7 @@ class KMLabel {
                 SUBMIT_RATING: 'Submit your rating',
                 CONVERSATION_RESOLVED: 'This conversation was resolved', //if conversation resolved through zendesk,
                 CONVERSATION_RATING_HEADING:
-                    'How satisfied are you with the conversation?',
+                    'How satisfied you are with the conversation?',
```
devashishmamgain, line 148: previous one is correct

AashishBhatt5: Done

devashishmamgain (dismissed):

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -2815,7 +2816,8 @@ var userOverride = {
                         .addClass('vis');
                 }
                 _this.setEmojiHoverText();
-                _this.configureRatingElements();
+                _this.configureStarsOrRatingElement();
+                // _this.configureRatingElements();
```
panwaranuj01, line 2820: remove unused code

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -3146,6 +3148,13 @@ var userOverride = {
                             $applozic('#mck-sidebox-ft').removeClass(
                                 'mck-restart-conv-banner km-mid-conv-csat'
                             );
+                            const stars = document.querySelectorAll(
```
panwaranuj01, line 3151: create a function inside the rating service and call it here

webplugin/js/app/components/rating-service.js
```diff
@@ -0,0 +1,78 @@
...
+        this.appOptions = appOptions;
+        this.isCsatRatingBase = widgetSettings.csatRatingBase
+            ? widgetSettings.csatRatingBase
+            : 3;
+        if (this.isCsatRatingBase == 5) {
+            this.starCsatHtmlEnable();
+        }
+    }
+    starCsatHtmlEnable = () => {
+        $applozic('.mck-smilies-inner').removeClass('vis').addClass('n-vis');
```
devashishmamgain, line 17: can we use direct javascript function and not use $applozic as we will be removing $applozic (jquery) usage at some point in order to reduce chat widget size
there is some existing function for add and remove class, that can be used

webplugin/js/app/components/rating-service.js
```diff
@@ -0,0 +1,78 @@
...
+        });
+    };
+
+    setStarsEffect = (rating) => {
+        const stars = document.querySelectorAll('.star-rating label');
+        let selectedRating = rating;
+
+        function highlightStars(index) {
+            for (let i = 0; i < stars.length; i++) {
+                stars[i].querySelector('svg path').style.fill =
```
devashishmamgain, line 45: better way would be to add class to the element, and let color codes be in the CSS files

webplugin/js/app/components/rating-service.js
```diff
@@ -0,0 +1,78 @@
...
+                    i <= index ? '#FFC045' : '#B3B3B3';
+            }
+        }
+        function removeHighlightFromStars() {
```
devashishmamgain, line 49: when using class, this can be done with one line without the need to loop through

devashishmamgain (changes requested):

devashishmamgain (approved):

panwaranuj01 (approved):

PR #1187, opened 2024-05-22
Added the dom purify for the message

panwaranuj01:
- Added the dom purify for the message

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 


webplugin/lib/js/dom-purify-3.1.4.min.js
```diff
@@ -0,0 +1,684 @@
+/*! @license DOMPurify 3.1.4 | (c) Cure53 and other contributors | Released under the Apache license 2.0 and Mozilla Public License 2.0 | github.com/cure53/DOMPurify/blob/3.1.4/LICENSE */
```
devashishmamgain, line 1: this file doesn't look like a minified version, please recheck

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -9378,6 +9378,11 @@ var userOverride = {
                 var attachmentBox = 'n-vis';
                 var kmAttchMsg = '';
                 let isUserMsg = true;
+
+                if (msg?.message) {
+                    msg.message = window.DOMPurify.sanitize(msg.message);
```
devashishmamgain, line 9383: what is the purpose of appending window. here
is it because of iFrame thing?

devashishmamgain (changes requested):

webplugin/js/app/mck-sidebox-1.0.js
panwaranuj01: window is referring to the iframe context here. 
This is a similar `window.Dom || Dom` but following the widget code convention

webplugin/lib/js/dom-purify-3.1.4.min.js
panwaranuj01: ok, it looks like I have done the formatting here  😄
will revert this

devashishmamgain (approved):

devashishmamgain (approved):

PR #1197, opened 2024-06-24
TDH-1003 -: Add the cache strategy to the plugin files

panwaranuj01:
- Add the cache strategy to the plugin files
- Some build have a default cache strategy 

### PR Checklist

### How was the code tested?

- Change the code and check after 3 min new bundle coming if there is new build is created

-

-

### Screenshot
- 

### DIFF
--- a/customHttp.yml
+++ b/customHttp.yml
@@ -1,21 +1,25 @@
-customHeaders:
-  - pattern: '**'
-    headers:
-      - key: 'Access-Control-Allow-Origin'
-        value: '*'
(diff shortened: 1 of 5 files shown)


devashishmamgain (approved):
@anujpanwar01 also check if there are any images/icons that need to have cache

panwaranuj01:
There are no images in the build folder images. Images directly served from the CDN and they already have 30 days cache period

devashishmamgain (approved):

PR #1198, opened 2024-06-24
TDH-1004 -: Load the files based on prop

panwaranuj01:
-

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 


webplugin/js/app/mck-app.js
```diff
@@ -311,6 +317,82 @@ function ApplozicSidebox() {
...
+                        async (file) => {
+                            promises.push(loadResourceAsync(file));
+                        }
+                    );
+                }
+            }
+
+            if (options.voiceNote) {
+                promises.push(
+                    loadResourceAsync(THIRD_PARTY_SCRIPTS.recorder.js)
```
devashishmamgain, line 379: if recorder and frvoice are both required together then better to have a single combined minified file for them

devashishmamgain (commented):
@panwaranuj01 does it work for case where the pre chat lead collection script is required for telephone numbers. do we wait until the script is loaded before showing the widget? or widget gets shown and the user is free to click on the chat icon

panwaranuj01:
> @panwaranuj01 does it work for case where the pre chat lead collection script is required for telephone numbers. do we wait until the script is loaded before showing the widget? or widget gets shown and the user is free to click on the chat icon

we are waiting till file not loaded fully.

webplugin/js/app/mck-app.js
panwaranuj01: That is good suggestion it will not make the extra call

panwaranuj01: done

devashishmamgain (dismissed):

devashishmamgain (approved):

devashishmamgain (approved):

PR #1204, opened 2024-07-22
TDH-1055 -: Parse only the widget related cookies

panwaranuj01:
- Parse only the widget-related cookies
- No need to decode the cookies because we are encoding the cookies

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 

### DIFF
--- a/webplugin/js/app/storage/cookie-service.js
+++ b/webplugin/js/app/storage/cookie-service.js
@@ -56,7 +56,7 @@ class KmCookieStorage extends KmStorage {
 
     getCookiePrefix = () => {
         const appOptions =
-            kmSessionStorage.getDataFromKmSession('appOptions') ||
+            appOptionSession.getPropertyDataFromSession('appOptions') ||
             applozic._globals;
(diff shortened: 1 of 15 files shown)


webplugin/js/app/storage/cookie-service.js
```diff
@@ -0,0 +1,99 @@
+class KmCookieStorage extends KmStorage {
+    constructor() {
+        super();
+    }
+
+    /* There is no need to decode the cookie because we are not encoding the cookie */
```
devashishmamgain, line 6: is it a new behaviour or earlier also we were not encoding? 
after this release, will there be any impact on the following case:

1. In the existing production release, User uses chat widget to send messages and some cookies are stored 
2. Next day, Kommunicate deploys this new version with the new implementation
3. Same user comes back and refresh the page and starting chatting.

Will it work seamlessly for the end user with the same cookies, same user id, chat history, etc?

panwaranuj01: We have never encoded the cookie.

devashishmamgain (approved):

PR #1207, opened 2024-08-02
Added the conversation metadata support

jaitens:
- Added the conversation metadata support

### PR Checklist

### How was the code tested?

- locally

### DIFF
--- a/webplugin/js/app/kommunicate-client.js
+++ b/webplugin/js/app/kommunicate-client.js
@@ -90,7 +90,7 @@ Kommunicate.client = {
             GROUP_CREATION_URL: window.kommunicate.IFRAME_OVERRIDES ? window.kommunicate.IFRAME_OVERRIDES.GROUP_CREATION_URL : parent.location.href,
             conversationMetadata:JSON.stringify(conversationDetail.metadata),
         };
-        if (kommunicateSettings.KM_TEST_PAGE) {
+        if (kommunicateSettings?.KM_TEST_PAGE) {
             groupMetadata.KM_TEST_PAGE = true;
         }
         typeof conversationDetail.teamId != 'undefined' &&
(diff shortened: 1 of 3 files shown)


webplugin/js/app/kommunicate-client.js
```diff
@@ -87,7 +90,9 @@ Kommunicate.client = {
             GROUP_CREATION_URL: window.kommunicate.IFRAME_OVERRIDES ? window.kommunicate.IFRAME_OVERRIDES.GROUP_CREATION_URL : parent.location.href,
             conversationMetadata:JSON.stringify(conversationDetail.metadata),
         };
-    
+        if (kommunicateSettings?.KM_TEST_PAGE) {
```
devashishmamgain, line 93: we can't have this code in the chat widget, figure out a way to add it through the website test page

devashishmamgain (changes requested):

webplugin/js/app/kommunicate-client.js
jaitens: we are passing KM_TEST_PAGE key from website test page in the defaultSetting and here we are capturing it.

devashishmamgain: i don't think this is the right way to do, there must be a way to achieve the same from website page javascript itself without touching chat widget code.
if required take help from @anujpanwar01

devashishmamgain (changes requested):

webplugin/js/app/kommunicate-client.js
devashishmamgain: this is weird, @anujpanwar01  is this not possible to pass the key: value pair as used in standard way

jaitens:
closing this PR. handle from website

panwaranuj01:
@jaitens Give the user access to update the default conversation metadata.
Create one function inside the `Kommunicate.js`.

jaitens:
need to create PR from development so closing this PR

PR #1211, opened 2024-08-20
Added the logic to load the iframe only if body is present

panwaranuj01:
-

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 

### DIFF
--- a/webplugin/plugin.js
+++ b/webplugin/plugin.js
@@ -107,7 +107,7 @@ var kmCustomIframe =
     '   display: none!important' +
     '} \n';
 
-isV1Script() ? injectJquery() : appendIframe();
+isV1Script() ? injectJquery() : appendIframeAfterBodyLoaded();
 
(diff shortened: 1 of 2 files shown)


webplugin/plugin.js
```diff
@@ -129,6 +129,28 @@ function removeElementFromHtmlById(elementIdArray) {
...
+        appendIframe();
+    } else if ('readyState' in window.parent.document) {
+        appendIframeAfterLoadedBody();
+    } else {
+        window.onload = function () {
+            appendIframe();
+        };
+    }
+
+    function appendIframeAfterLoadedBody() {
```
devashishmamgain, line 144: rename this function to avoid confusion with appendIframeAfterBodyLoaded

took me sometime to realize that these are different names

panwaranuj01: changed the name

devashishmamgain (approved):

PR #1220, opened 2024-10-04
added support for esc button on image preview

thakaresakshi121:
- when we click on ESC button  "image preview" should be closed.

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -3569,6 +3569,14 @@ const firstVisibleMsg = {
                 ).style.display = 'none';
             };
 
+            window.onkeyup= function(event){
+              if(event.keyCode==27) {
+                parent.document.getElementById(
+                    'km-fullscreen-image-modal'
+                ).style.display = 'none';
+              }
+            }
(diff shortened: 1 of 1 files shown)


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -3569,6 +3569,14 @@ const firstVisibleMsg = {
...
             };
 
+            window.onkeyup= function(event){
+              if(event.keyCode==27) {
+                parent.document.getElementById(
+                    'km-fullscreen-image-modal'
+                ).style.display = 'none';
+              }
+            }
+
```
panwaranuj01, line 3579: @thakaresakshi121 If any user is using window.onkeyup in their code, it will either overwrite their existing listener or remove our listener.

Try to use the event in the element instead of the window object.

panwaranuj01 (changes requested):

panwaranuj01 (changes requested):
It will be better to remove the event listener when modal will closed

panwaranuj01 (approved):

PR #1228, opened 2024-10-15
TDH-1330 -: Help/Not Helpful

panwaranuj01:
- Help/Not Helpful
- added the core changes later we can change the design

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 

### DIFF
--- a/webplugin/js/app/components/answer-feedback-service.js
+++ b/webplugin/js/app/components/answer-feedback-service.js
@@ -1,0 +1,160 @@
+class AnswerFeedback {
+    constructor(options) {
+        this.options = options;
+        this.msgMap = {};
+        this.msgKeyMap = [];
+    }
+
+    /**
+     * @param {{ key: string | number; }} data
+     */
+    set msgMapData(data) {
+        const msgKey = data.msg.key;
+
+        this.msgMap[msgKey] = data;
+        this.msgKeyMap.push(msgKey);
+    }
+
+    handleOnFeedbackClick = (feedback, data) => {
+        const { payload, method = 'POST' } = this.getPayload(feedback, data);
+
+        mckUtils.ajax({
+            type: method,
+            url: `${Kommunicate.getBaseUrl()}/rest/ws/answer-feedback`,
+            headers: {
+                'x-authorization': window.Applozic.ALApiService.AUTH_TOKEN,
+            },
+            data: JSON.stringify({ payload }),
+            global: false,
+            contentType: 'application/json',
+            success: function (result) {
+                console.log(result);
+                if (result.status == 'success') {
+                } else {
+                }
+            },
+            error: function (data) {
+                console.error(data);
+            },
+        });
+    };
+
+    getPayload = (feedback, data) => {
+        const {
+            msg: { metadata, key },
+            assigneeKey,
+        } = data;
+
+        const answer = this.msgMap[key];
+        let question = answer;
+
+        // find the user message
+        while (
+            !metadata.KM_CONTEXT_QUESTION &&
+            question?.floatWhere !== 'mck-msg-right'
+        ) {
+            question = question.previousSibling;
+        }
+
+        const answerText = answer.msg.message;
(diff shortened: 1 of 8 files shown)


panwaranuj01:
Ignore the css for now because later that will be changed

webplugin/js/app/components/answer-feedback-service.js
```diff
@@ -0,0 +1,210 @@
...
+            assigneeKey,
+        } = data;
+
+        const answer = this.msgMap[key];
+        let question = answer;
+
+        // find the user message
+        while (
+            !metadata.KM_CONTEXT_QUESTION &&
+            question?.floatWhere !== 'mck-msg-right'
```
devashishmamgain, line 87: this will be based on RTL setting?

webplugin/js/app/components/answer-feedback-service.js
```diff
@@ -0,0 +1,210 @@
...
+        };
+    };
+
+    getActiveClass = (msg, feedback) => {
+        const { KM_ANSWER_FEEDBACK } = msg.metadata;
+
+        return KM_ANSWER_FEEDBACK == feedback ? 'active-feedback' : '';
+    };
+
+    helpFullOnClick = (data) => {
```
devashishmamgain, line 121: helpful
and not helpfull

webplugin/js/app/components/answer-feedback-service.js
```diff
@@ -0,0 +1,210 @@
...
+
+        if (
+            currentUser.role !==
+            KommunicateConstants.GROUP_ROLE.MODERATOR_OR_BOT
+        ) {
+            return false;
+        }
+
+        return (
+            floatWhere === 'mck-msg-left' &&
```
devashishmamgain, line 158: check if it will be impacted by RTL setting, figure out some other way to distinguish it without specifying left or right

webplugin/js/app/components/answer-feedback-service.js
```diff
@@ -0,0 +1,210 @@
...
+    };
+
+    attachEventListeners = (data) => {
+        const { msg } = data;
+
+        const msgContainer = document.querySelector(
+            `[data-msgKey='${msg.key}']`
+        );
+
+        const helpfulButton = msgContainer.querySelector(
```
devashishmamgain, line 171: see if something like this simplify it
const setFeedback = (isHelpful) => {
    helpfulButton.classList.toggle('active-feedback', isHelpful);
    notHelpfulButton.classList.toggle('active-feedback', !isHelpful);
    isHelpful ? this.helpFullOnClick(data) : this.notHelpFullOnClick(data);
};

const helpfulButton = msgContainer.querySelector('.answer-feedback-helpful');
const notHelpfulButton = msgContainer.querySelector('.answer-feedback-not-helpful');

helpfulButton.addEventListener('click', () => setFeedback(true));
notHelpfulButton.addEventListener('click', () => setFeedback(false));

webplugin/js/app/km-widget-custom-events.js
```diff
@@ -215,4 +215,20 @@ var eventMapping = {
         },
         eventFunction: null,
     },
+    onHelpFulClick: {
```
devashishmamgain, line 218: do we need 2 separate events? or we can have just one event, and receive the value on that event with 0 or 1
in future, if we add more capabilities say: Great Answer, Loved it, etc then this code will not be scalable

devashishmamgain (changes requested):

webplugin/js/app/components/answer-feedback-service.js
panwaranuj01: we are changing the class if it's rtl or ltr. This will work

webplugin/js/app/components/answer-feedback-service.js
panwaranuj01: this will not impact the rtl

devashishmamgain (approved):

PR #1229, opened 2024-10-15
  Summary Text in widget [FWH-2540]

deepaktiwari03:
- Fix summary in contact box issue in widget conversation summary bug

### PR Checklist

### How was the code tested?

- Tested by generating the summary

### Screenshot
- 

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -12459,7 +12459,7 @@ const firstVisibleMsg = {
                 message,
                 contact
             ) {
-                if (message?.metadata?.KM_SUMMARY === 'true') {
+                if (message?.metadata?.category === 'HIDDEN') {
                     return;
                 }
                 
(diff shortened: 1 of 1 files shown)


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -12459,6 +12459,10 @@ const firstVisibleMsg = {
                 message,
                 contact
             ) {
+                if (message?.metadata?.KM_SUMMARY === 'true') {
+                    return;
+                }
+
```
panwaranuj01, line 12465: @deepaktiwari03 when are doing the summarize we are adding the `category: hidden` in that message from backend?

deepaktiwari03: Yes, we are receiving that but still message was message in the message box list.

panwaranuj01: Then we can the add the check for `category: hidden` here instead of the KM_SUMMARY;
refer the visibleMessage function

deepaktiwari03: sure. will add it

deepaktiwari03: added please check now

PR #1233, opened 2024-10-21
OPS-785: deployment script for firebase

adarshmishra:
- deployment script for firebase. 
- NOTE: Project is currently selected by default. Ops team will create pipeline and pass project and firebase credentials dynamically. 

### PR Checklist

### How was the code tested?

- https://widget-release.kommunicate.io/  Currently hosted on Firbase

-

-

### Screenshot
- 

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -8,11 +8,24 @@
     ],
     "headers": [
       {
-        "source": "**/*.html",
+        "source": "**",
         "headers": [
           {
             "key": "Access-Control-Allow-Origin",
             "value": "*"
+          }
+        ]
+      },
+      {
+        "source": "/v2/kommunicate.app",
+        "headers": [
+          {
+            "key": "Content-Type",
+            "value": "text/javascript"
+          },
+          {
+            "key": "Cache-Control",
+            "value": "public, max-age=180"
           }
         ]
       },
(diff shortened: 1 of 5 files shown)


devashishmamgain (changes requested):
check for cache period setting https://github.com/Kommunicate-io/Kommunicate-Web-SDK/blob/master/customHttp.yml

firebase.json
```diff
@@ -0,0 +1,42 @@
...
+    "headers": [
+      {
+        "source": "**/*.html",
+        "headers": [
+          {
+            "key": "Access-Control-Allow-Origin",
+            "value": "*"
+          }
+        ]
+      },
```
panwaranuj01, line 18: There is one file, `kommunicate.app`, which is the initial point of the script basically, we are giving that URL in the customer script, and it does not follow any file convention these regex will not match that file

panwaranuj01 (changes requested):

panwaranuj01 (commented):
And we also have to add this is header for only this file kommunicate.app
```
  - pattern: "/v2/kommunicate.app"
    headers:
      - key: "Content-Type"
        value: "text/javascript"
      - key: "Cache-Control"
        value: "public, max-age=180"

```

firebase.json
adarshmishra: added

adarshmishra:
> check for cache period setting https://github.com/Kommunicate-io/Kommunicate-Web-SDK/blob/master/customHttp.yml

Converted according to above file, @devashishmamgain @anujpanwar01

panwaranuj01 (approved):

PR #1240, opened 2024-10-29
CSAT is working as expected

swrjnnd:
CSAT should not be reappear everytime

**Previous Behavior:**
- The CSAT (Customer Satisfaction) rating prompt appears every time we resolve a conversation. The prompt remains active until the customer provides a rating.

**Current Behavior:**
- The CSAT prompt should close and not reappear under the following conditions:
  1. When we restart the conversation and then resolve it.
  2. If we do not close the CSAT using the button and instead refresh the page. This is necessary because if the customer does not close the CSAT or provide a rating, we want to ensure it reappears as needed.

### PR Checklist

- [ ] I have tested it locally, and all functionalities are working correctly.
- [ ] I have compared it with design mocks, and all elements match.
- [ ] I have tested it in Internet Explorer.

### Code Testing Details
- The code has been thoroughly tested locally to ensure all functionalities are operating as expected.

### DIFF
--- a/webplugin/js/app/conversation/km-conversation-service.js
+++ b/webplugin/js/app/conversation/km-conversation-service.js
@@ -20,11 +20,18 @@ Kommunicate.conversation = {
             );
             return;
         }
+
         var conversationDetail = data && data.groupFeeds[0];
+        const isConversationRated = kmLocalStorage.getItemFromLocalStorage(
+            conversationDetail.id
+        );
+
         KommunicateUI.showClosedConversationBanner(
-            Kommunicate.conversationHelper.isConversationClosed(
-                conversationDetail
-            )
+            isConversationRated.isConversationClosed
+                ? false
+                : Kommunicate.conversationHelper.isConversationClosed(
+                      conversationDetail
+                  )
         );
     },
 };
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -16,6 +16,7 @@ var WAITING_QUEUE = [];
 var AVAILABLE_VOICES_FOR_TTS = new Array();
 var KM_ATTACHMENT_V2_SUPPORTED_MIME_TYPES = ['application', 'text', 'image'];
 const DEFAULT_TEAM_NAME = ['Default Team', 'Default'];
+var CHAT_GROUP_ID = "";
 var userOverride = {
     voiceOutput: true,
 };
@@ -414,6 +415,7 @@ const firstVisibleMsg = {
         var MCK_GROUPMAXSIZE = appOptions.maxGroupSize;
         var MCK_ON_TAB_CLICKED = function (event) {
             console.log('In on_tab_clicked', event);
+            CHAT_GROUP_ID = event && event.tabId;
             const details = event && event.data && event.data.groupDetails;
             if (details) {
                 const assignee =
@@ -5046,6 +5048,10 @@ const firstVisibleMsg = {
                     'km-csat-close-button'
                 ).onclick = function (e) {
                     e.preventDefault();
+                    kmLocalStorage.setItemToLocalStorage(
+                        CHAT_GROUP_ID, { isConversationClosed : true }
+                    )
+
                     KommunicateUI.showClosedConversationBanner(false);
                 };
 
@@ -6445,6 +6451,11 @@ const firstVisibleMsg = {
                     data: w.JSON.stringify(messagePxy),
                     contentType: 'application/json',
                     success: function (data) {
+                        const {groupId} = messagePxy;
+
+                        kmLocalStorage.setItemToLocalStorage(
+                            groupId, { isConversationClosed :false }
+                        )
                         if (kommunicate._globals.zendeskChatSdkKey) {
                             zendeskChatService.handleUserMessage(messagePxy);
                         }


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -5046,6 +5048,10 @@ const firstVisibleMsg = {
                     'km-csat-close-button'
                 ).onclick = function (e) {
                     e.preventDefault();
+                    kmLocalStorage.setItemToLocalStorage(
+                        CHAT_GROUP_ID, { isConversationClosed : true }
```
devashishmamgain, line 5052: is the 'conversation' getting closed today, or its simply about user closing the CSAT modal?

devashishmamgain (changes requested):

webplugin/js/app/conversation/km-conversation-service.js
```diff
@@ -20,11 +20,18 @@ Kommunicate.conversation = {
             );
             return;
         }
+
         var conversationDetail = data && data.groupFeeds[0];
+        const isConversationRated = kmLocalStorage.getItemFromLocalStorage(
+            conversationDetail.id
+        );
```
panwaranuj01, line 27: instead of localStorage use the session storage

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -5046,6 +5048,10 @@ const firstVisibleMsg = {
                     'km-csat-close-button'
                 ).onclick = function (e) {
                     e.preventDefault();
+                    kmLocalStorage.setItemToLocalStorage(
+                        CHAT_GROUP_ID, { isCSATmodelClosed : true }
```
devashishmamgain, line 5052: M as caps

devashishmamgain (changes requested):

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -414,6 +415,7 @@ const firstVisibleMsg = {
         var MCK_GROUPMAXSIZE = appOptions.maxGroupSize;
         var MCK_ON_TAB_CLICKED = function (event) {
             console.log('In on_tab_clicked', event);
+            CHAT_GROUP_ID = event && event.tabId;
```
panwaranuj01, line 418: We already maintaining the current active group data please use that

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -5046,6 +5048,10 @@ const firstVisibleMsg = {
                     'km-csat-close-button'
                 ).onclick = function (e) {
                     e.preventDefault();
+                    kmLocalStorage.setItemToLocalStorage(
+                        CHAT_GROUP_ID, { isCSATmodelClosed : true }
+                    )
```
panwaranuj01, line 5053: It should be some like this 

```
{
[groupId]: true/false
}
```

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -6445,6 +6451,11 @@ const firstVisibleMsg = {
                     data: w.JSON.stringify(messagePxy),
                     contentType: 'application/json',
                     success: function (data) {
+                        const {groupId} = messagePxy;
+
+                        kmLocalStorage.setItemToLocalStorage(
+                            groupId, { isCSATmodelClosed :false }
+                        )
```
panwaranuj01, line 6458: If user opens the old group then it will be break; 
try to add this real-time message

panwaranuj01 (changes requested):

webplugin/js/app/conversation/km-conversation-service.js
swrjnnd: If we use session storage instead of local storage, there will be one issue: once the user closes the CSAT feedback, it will not appear again until the page is refreshed or reopened. When the user closes and reopens the page, the CSAT feedback will reappear once, as session storage is cleared at the end of each session. so that's why i am using the local storage instead of session storage.

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -6445,6 +6450,11 @@ const firstVisibleMsg = {
                     data: w.JSON.stringify(messagePxy),
                     contentType: 'application/json',
                     success: function (data) {
+                        const {groupId} = messagePxy;
+
```
devashishmamgain, line 6454: are you expecting only one group for the customer and that too always opened in that tab?

devashishmamgain (changes requested):

panwaranuj01:
@coderabbitai review

devashishmamgain (changes requested):
will this work if say user has 2 different conversations and opens both of them, then comes back to the main list

swrjnnd:
> will this work if say user has 2 different conversations and opens both of them, then comes back to the main list

Yes sir it will work i tested with this case also it is working fine

devashishmamgain:
> > will this work if say user has 2 different conversations and opens both of them, then comes back to the main list
> 
> Yes sir it will work i tested with this case also it is working fine

so on logout, all the multiple groups group id local storage is getting cleared?

devashishmamgain (approved):

panwaranuj01 (approved):

PR #1242, opened 2024-10-29
Gcp ci firebase

GhiyasKM:
-Have added rewrite and redirect in firebase.json 
-Also have improved cloudbuild.yaml file
### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -2,8 +2,8 @@ options:
   logging: CLOUD_LOGGING_ONLY
 
 substitutions:
-  _BUILD_ENV: "test"  # Can be set per environment (test, release)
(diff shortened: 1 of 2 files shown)


cloudbuild.yaml
```diff
@@ -30,83 +37,133 @@ steps:
...
                 ]
               }
+            ],
+            "rewrites": [
+              {
+                "source": "/<*>",
+                "destination": "/index.html",
+                "type": "404"
+              },
+              {
```
adarshmishra, line 106: @GhiyasKM as discussed, simplified the script and use exiting firbebase.json file

adarshmishra (approved):

PR #1248, opened 2024-11-05
Fix conversation message on message 

deepaktiwari03:
- Add hard check to disable any conversation summary summary message in widget conversation list and message list

### PR Checklist

### How was the code tested?

- Tested locally

### Issue
- Message received after talk to human button is click doesn't contain hidden in metadata. Now, we are returning if it contain KM_SUMMARY so that no message which contain this field is added to the widget. 

### Screenshot
- after fix
Screenshot 2024-11-05 at 1 35 29 PM

- before fix 
Screenshot 2024-11-05 at 1 20 13 PM

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -9482,6 +9482,14 @@ const firstVisibleMsg = {
                 }
 
                 if (
+                    msg &&
+                    msg.metadata &&
+                    msg.metadata.hasOwnProperty('KM_SUMMARY')
+                ) {
+                    return;
+                }
+
+                if (
                     $applozic('#mck-message-cell .' + msg.key).length > 0 &&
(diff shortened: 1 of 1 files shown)


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -9481,6 +9481,14 @@ const firstVisibleMsg = {
                     }
                 }
 
+                if (
+                    msg &&
```
devashishmamgain, line 9485: what is the purpose of Kommunicate.visibleMessage ? is that meant for this purpose and check needs to go there?

deepaktiwari03: I have added this in addMessage. Not sure why it is showing here. This variable is defined in start of code

devashishmamgain (approved):

panwaranuj01 (approved):

PR #1269, opened 2024-12-12
CUS-730 Chat widget should open again if entered password is wrong and page is refreshed.

jaitens:
- After entering agent Email it ask for password if you enter wrong password and refresh the page Widget UI breaks.

### PR Checklist

### How was the code tested?

- Locally

- The issue arises because we are storing LeadCollection: true in the cookies. Upon page reload, we check if LeadCollection is set to true, and if so, we avoid asking for the user's details again, automatically logging them in with the previously stored email. However, since a password is also required, this causes the UI to break when an invalid password is entered. To resolve this, we now delete the LeadCollection value from the cookies whenever an "Invalid Password" error occurs.

### Screenshot
- 

## Summary by CodeRabbit

- **New Features**
  - Introduced a `robots.txt` file to manage web crawler access.
  - Added a new rewrite rule in `firebase.json` to serve the `robots.txt` file correctly.
  - Enhanced the build process to include the generation of the `robots.txt` file.

- **Bug Fixes**
  - Improved dependency management during the build process for Node.js applications.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -7,13 +7,22 @@ substitutions:
   _FIREBASE_SITE_ID: "${_BUILD_ENV}_site_id"  # Placeholder for Firebase site ID
 
 steps:
-  # Step 1: Install dependencies
+ # Step 1: Install Node.js dependencies
   - name: 'gcr.io/cloud-builders/npm'
-    args: ['install']
+    entrypoint: 'sh'
+    args:
+      - '-c'
+      - |
+        echo clean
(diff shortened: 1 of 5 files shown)


devashishmamgain (dismissed):

devashishmamgain (commented):
@jaitens lets use a common function that takes care of all stored values

devashishmamgain (approved):

panwaranuj01 (approved):

PR #1277, opened 2024-12-30
CUS-766 Disable handover button till api response is received after click

jaitens:
- Disable handover button till api response is received after talk to agent button is clicked.

### PR Checklist

### How was the code tested?

- locally

- Talk to human button was not getting disabled when api call is happening.

### Screenshot
- 

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -5175,6 +5175,8 @@ const firstVisibleMsg = {
 
                 $applozic(d).on('click', '#km-talk-to-human', function (e) {
                     e.preventDefault();
+                    var $button = $applozic(this);
+                    $button.prop('disabled', true);
 
                     window.Applozic.ALApiService.ajax({
                         type: 'PATCH',
@@ -5201,8 +5203,10 @@ const firstVisibleMsg = {
                                         );
                                 }
                             }
+                            $button.prop('disabled', false); // enable button
                         },
                         error: function (data) {
+                            $button.prop('disabled', false); // enable button
                             console.error(data);
                         },
                     });


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -5175,6 +5175,8 @@ const firstVisibleMsg = {
 
                 $applozic(d).on('click', '#km-talk-to-human', function (e) {
                     e.preventDefault();
+                    var $button = $applozic(this);
```
panwaranuj01, line 5178: @jaitens Let's use the javascript instead of the jQuery because we are trying to remove the jquery

panwaranuj01 (changes requested):

webplugin/js/app/mck-sidebox-1.0.js
jaitens: done

webplugin/js/app/km-nav-bar.js
```diff
@@ -34,6 +34,7 @@ function KmNavBar(mckMsgLayout) {
 
     _this.modifyTalkToHumanVisibility = function (isAssignedToHuman) {
         if (isAssignedToHuman) {
+            document.getElementById("km-talk-to-human").disabled = false;
```
devashishmamgain, line 37: can this throw undefined error ever if there is no element on UI?

jaitens: it will not throw any error because modifyTalkToHumanVisibility get called when user clicks on the button so if user is able to click on the button that means element is available and it will not throw error.

For safety adding optional chaining.

devashishmamgain (dismissed):

webplugin/js/app/km-nav-bar.js
```diff
@@ -34,6 +34,7 @@ function KmNavBar(mckMsgLayout) {
 
     _this.modifyTalkToHumanVisibility = function (isAssignedToHuman) {
         if (isAssignedToHuman) {
+            document.getElementById("km-talk-to-human")?.disabled = false;
```
panwaranuj01, line 37: @jaitens This will throw an error if km-talk-to-human is not there

panwaranuj01 (changes requested):

panwaranuj01 (approved):

PR #1290, opened 2025-02-10
Handle preview if url is protected or giving cloud flare error

jaitens:
- Preventing preview if url is protected or giving cloud flare error

### PR Checklist

### How was the code tested?

-

-

- In api server we are using a library link-preview-js to get data for preview. So if any website having captcha or if there any website is using security service to protect itself from online attacks then also this library gives success response. And it shows whatever data response it gets from the API. So to prevent this added some checks and if it not passes those checks then it will not show preview.

Here are some website which are having captcha or security services.
- https://www.saveonfoods.com/deals
Screenshot 2025-02-10 at 4 04 29 PM

- https://securityheaders.com/
Screenshot 2025-02-10 at 4 05 09 PM

### Screenshot
- 

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved link preview reliability by adding error handling for missing titles and descriptions, ensuring only valid content is displayed.
- **Style**
	- Enhanced visual spacing in message components by updating padding for a cleaner layout.

### DIFF
--- a/webplugin/js/app/kommunicate-ui.js
+++ b/webplugin/js/app/kommunicate-ui.js
@@ -139,15 +139,35 @@ KommunicateUI = {
             type: 'GET',
             global: false,
             success: function (result) {
+                const { title, description } = result.data;
+                // Check if the title contains Cloudflare or security challenge related words
+                if (
+                    title &&
+                    title.includes('Attention Required!')
+                ) {
+                    throw Error("Cloudflare or security block detected. No preview available.") // Skip rendering the preview or return a default message
+                }
+
+                // Check for other possible issues with the preview data (e.g., missing meta information)
+                if (!title || !description) {
+                    throw Error("Missing metadata for preview. No preview available.")
+                }
                 if (result) {
                     var images = result.data.images;
-                    result.data.images = images.length ? KommunicateUI.checkSvgHasChildren(images) : [];
+                    result.data.images = images.length
+                        ? KommunicateUI.checkSvgHasChildren(images)
+                        : [];
 
                     // this happens when the link gets redirected
-                    if (result.data.title === "ERROR: The request could not be satisfied") return;
+                    if (
+                        result.data.title ===
+                        'ERROR: The request could not be satisfied'
+                    )
+                        return;
 
                     var previewTemplate = kommunicate.markup.getLinkPreviewTemplate(
-                        result, isMckRightMsg
+                        result,
+                        isMckRightMsg
                     );
                     callback(previewTemplate);
                 }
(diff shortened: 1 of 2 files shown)


webplugin/js/app/kommunicate-ui.js
```diff
@@ -139,15 +139,35 @@ KommunicateUI = {
             type: 'GET',
             global: false,
             success: function (result) {
+                const { title, description } = result.data;
```
panwaranuj01, line 142: @jaitens this destructuring can throw an error if api fails, handle it properly.

webplugin/js/app/kommunicate-ui.js
```diff
@@ -139,15 +139,35 @@ KommunicateUI = {
...
+                // Check if the title contains Cloudflare or security challenge related words
```
panwaranuj01, line 143: no need these comment or maybe keep it shorter

panwaranuj01 (changes requested):

webplugin/js/app/kommunicate-ui.js
```diff
@@ -139,15 +139,35 @@ KommunicateUI = {
...
+                if (
+                    title &&
+                    title.includes('Attention Required!')
+                ) {
+                    throw Error("Cloudflare or security block detected. No preview available.") // Skip rendering the preview or return a default message
+                }
+
+                // Check for other possible issues with the preview data (e.g., missing meta information)
+                if (!title || !description) {
+                    throw Error("Missing metadata for preview. No preview available.")
```
panwaranuj01, line 153: do we really need to throw the error, or can we just return?

webplugin/js/app/kommunicate-ui.js
```diff
@@ -139,15 +139,35 @@ KommunicateUI = {
             type: 'GET',
             global: false,
             success: function (result) {
+                const { title, description } = result.data;
+                // Check if the title contains Cloudflare or security challenge related words
+                if (
+                    title &&
+                    title.includes('Attention Required!')
```
devashishmamgain, line 146: what does other players similar to Cloudflare shows?
Is this a standard practice of checking text part for the previous thing? or there are any better and efficient ways to do it?
can you check with Twiter, LinkedIN and other places see what does they show in preview with such urls and how does they handle it?

devashishmamgain (changes requested):

webplugin/js/app/kommunicate-ui.js
jaitens: Checked in linkedin they are able to get the preview   
Screenshot 2025-02-11 at 1 33 17 PM

And in whatsapp it show the URL only in the preview
Screenshot 2025-02-11 at 1 37 49 PM

The library we are using to get data for preview is giving success response. It is not throwing any error so it might not possible to say when the URL is protected thats why checking with the text. I tried searching for different library to get preview data but not getting anything which can resolve this issue.

devashishmamgain (approved):

panwaranuj01 (approved):

PR #1299, opened 2025-02-26
Tdh 1776 Restrict Human Handover

AashishBhatt5:
- Instead of Assignee change api now we will use send message api. Because assignee change api will always change assignee but if bot to human handover is off when agents are away/offline in that case we want to show away message too

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -5196,13 +5196,11 @@ const firstVisibleMsg = {
                         key: mckUtils.randomId(),
                         metadata: {
                             TALK_TO_HUMAN: true,
-                            KM_CHAT_CONTEXT: JSON.stringify(
-                                _this.getChatContext(messagePxy)
-                            ),
                         },
                         groupId: CURRENT_GROUP_DATA.tabId,
                         source: 1,
                     };
+                    messagePxy.metadata.KM_CHAT_CONTEXT = JSON.stringify(_this.getChatContext(messagePxy));
                     _this.sendMessage(messagePxy);
                     const lastMessageBeforeSend = $applozic(
                         "#mck-message-cell .mck-message-inner div[name='message']:last-child"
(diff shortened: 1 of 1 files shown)


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -5188,38 +5189,29 @@ const firstVisibleMsg = {
...
+                            KM_CHAT_CONTEXT: JSON.stringify(
+                                _this.getChatContext(messagePxy)
+                            ),
                         },
-                    });
+                        groupId: CURRENT_GROUP_DATA.tabId,
+                        source: 1,
+                    };
+                    _this.sendMessage(messagePxy);
+                    var lastMessageBeforeSend = $applozic(
```
panwaranuj01, line 5207: use const

panwaranuj01, line 5207: use the js instead of jquery

panwaranuj01 (changes requested):

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -13116,6 +13111,17 @@ const firstVisibleMsg = {
...
+                    let talkToHumanButton = document.getElementById(
+                        'km-talk-to-human'
+                    );
+                    if (
+                        talkToHumanButton?.disabled &&
+                        message?.metadata?.KM_ASSIGN_TO === ''
+                    ) {
+                        talkToHumanButton.disabled = false;
+                    }
+                }
```
panwaranuj01, line 13124: It can be added in the onMessage function

webplugin/js/app/mck-sidebox-1.0.js
AashishBhatt5: Done

webplugin/js/app/mck-sidebox-1.0.js
AashishBhatt5: Done

webplugin/js/app/mck-sidebox-1.0.js
AashishBhatt5: done

panwaranuj01 (approved):

PR #1301, opened 2025-02-28
Revert app_id changes

jaitens:
-

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 

## Summary by CodeRabbit

- **Chores**
  - Updated the logic for error handling related to application settings, ensuring clearer conditions for `appId` and `widgetPlatformUrl`. Users are still instructed to replace "APP_ID" with their actual App ID if necessary.

### DIFF
--- a/webplugin/js/app/mck-app.js
+++ b/webplugin/js/app/mck-app.js
@@ -737,7 +737,7 @@ function ApplozicSidebox() {
         applozic._globals.widgetPlatformUrl &&
             (data.widgetPlatformUrl = applozic._globals.widgetPlatformUrl);
         // NOTE: Don't pass applozic._globals as it is in data field of ajax call, pass only the fields which are required for this API call.
-        if (data?.appId === "APP_ID") {
+        if (!data.widgetPlatformUrl && (!data.appId || data.appId === "APP_ID" || data.appId === "")) {
             console.error("Please replace 'APP_ID' with your App ID in the Kommunicate plugin settings.");
             return;
         }
--- a/webplugin/js/app/mck-app.js
+++ b/webplugin/js/app/mck-app.js
@@ -737,6 +737,11 @@ function ApplozicSidebox() {
         applozic._globals.widgetPlatformUrl &&
             (data.widgetPlatformUrl = applozic._globals.widgetPlatformUrl);
         // NOTE: Don't pass applozic._globals as it is in data field of ajax call, pass only the fields which are required for this API call.
+        if (data?.appId === "APP_ID") {
+            console.error("Please replace 'APP_ID' with your App ID in the Kommunicate plugin settings.");
+            return;
+        }
+        
         var url =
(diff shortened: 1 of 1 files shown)


panwaranuj01 (approved):

devashishmamgain (changes requested):
@jaitens also figure out why was this code added initially, it must have been added to fix some issue, the final fix should solve both issues old and new one.

jaitens:
> @jaitens also figure out why was this code added initially, it must have been added to fix some issue, the final fix should solve both issues old and new one.

https://github.com/Kommunicate-io/Kommunicate-Web-SDK/pull/1271

in this PR it is written 
Someone has copy pasted APP_ID as it is from docs page and that making unnecessary hits to the server with appId=APP_ID

so adding check to early return if appId == APP_ID

devashishmamgain:
> > @jaitens also figure out why was this code added initially, it must have been added to fix some issue, the final fix should solve both issues old and new one.
> 
> #1271
> 
> in this PR it is written Someone has copy pasted APP_ID as it is from docs page and that making unnecessary hits to the server with appId=APP_ID
> 
> so adding check to early return if appId == APP_ID

yes there are millions of errors coming on server for this, the fix should be to support both this issue as well as Shopify issue.
We can't bring back Problem A to fix Problem B. Figure out a way to have both A and B fixed together.

jaitens:
> > > @jaitens also figure out why was this code added initially, it must have been added to fix some issue, the final fix should solve both issues old and new one.
> > 
> > 
> > #1271
> > in this PR it is written Someone has copy pasted APP_ID as it is from docs page and that making unnecessary hits to the server with appId=APP_ID
> > so adding check to early return if appId == APP_ID
> 
> yes there are millions of errors coming on server for this, the fix should be to support both this issue as well as Shopify issue. We can't bring back Problem A to fix Problem B. Figure out a way to have both A and B fixed together.

yes added check to handle both the issue. please review

webplugin/js/app/mck-app.js
```diff
@@ -737,10 +737,11 @@ function ApplozicSidebox() {
         applozic._globals.widgetPlatformUrl &&
             (data.widgetPlatformUrl = applozic._globals.widgetPlatformUrl);
         // NOTE: Don't pass applozic._globals as it is in data field of ajax call, pass only the fields which are required for this API call.
-        if (!data.appId || data.appId === "" || data.appId === "APP_ID") {
+        if (data?.appId === "APP_ID") {
```
panwaranuj01, line 740: @jaitens We can add to check if the widgetPlatformUrl is provided, then don't check for the appId, otherwise check for the appId validation

panwaranuj01 (changes requested):

panwaranuj01 (dismissed):

devashishmamgain (approved):

PR #1315, opened 2025-04-23
CUS-464 When we copy paste message in chat widget the send option doesn't appear needs to enter space to enable

jaitens:
- When we copy paste message in chat widget the send option doesn't appear needs to enter space to enable

### PR Checklist

### How was the code tested?

-

-

- keyup event was used due which send button was not showing so added input event

### Screenshot
- 
Screenshot (578)

## Summary by CodeRabbit

- **Refactor**
  - Updated the message input field to respond to all input changes, including paste and cut actions, for improved responsiveness.

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -3907,7 +3907,7 @@ const firstVisibleMsg = {
                     mckMessageLayout.addGroupsToGroupSearchList();
                 });
 
-                mck_text_box.addEventListener('keyup', function () {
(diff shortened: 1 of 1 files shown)


panwaranuj01 (commented):
why we are changing to input event? 
When keyup event wont trigger?

jaitens:
> why we are changing to input event? When keyup event wont trigger?

@anujpanwar01 when we copy paste from the option we get after right click (you can see the screenshot attached in desc. ) then keyup event wont fire thats why using input event

panwaranuj01 (approved):

PR #1317, opened 2025-05-19
Fix agentic bugs

thakaresakshi121:
- fixed issue with copy pasting msg with CSS . now only text will be pasted not CSS will be pasted .
-  fixed issue with answer feedback position 

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 

### DIFF
--- a/webplugin/scss/components/_km-message-area.scss
+++ b/webplugin/scss/components/_km-message-area.scss
@@ -37,8 +37,7 @@
 }
 .km-answer-feedback {
     position: absolute;
-    bottom: 0;
-    right: 25%;
+    bottom: 5px;
     display: flex !important;
     justify-content: flex-end;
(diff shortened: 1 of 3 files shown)


webplugin/scss/components/_km-message-area.scss
```diff
@@ -37,8 +37,7 @@
 }
 .km-answer-feedback {
     position: absolute;
-    bottom: 0;
-    right: 25%;
+    bottom: 5px;
```
panwaranuj01, line 40: Is it will work on the smaller screen?

thakaresakshi121: yes tested on both screen

panwaranuj01 (approved):

PR #1320, opened 2025-05-21
Tdh 1851

panwaranuj01:
-

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 


webplugin/js/app/kommunicate.custom.theme.js
```diff
@@ -102,4 +103,10 @@ function KmCustomTheme() {
             messageBoxTop[i].style.backgroundColor = WIDGET_SETTINGS.primaryColor;
         }
     };
+
+    _this.fillSvgsTheme = function () {
+        document.querySelectorAll("path[fill=':primary']").forEach((path) => {
+            path.setAttribute('fill', WIDGET_SETTINGS.primaryColor);
+        });
+    };
```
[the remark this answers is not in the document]
panwaranuj01: default color will be there

webplugin/js/app/voice/index.js
```diff
@@ -0,0 +1,70 @@
+class Voice {
+    _VOICE_PLATFORM_API_URL = 'https://api.elevenlabs.io';
+    _VOICE_PLATFORM_API_KEY = 'sk_5af5c862dd8c87860ece549378c67628233440a018ac2e12';
```
devashishmamgain, line 3: can this be a problem as this is on customer's website so we don't have control of restricting in any domain?
better to download this key from server, so that if we ever to have change this api key, we can change it from server

webplugin/template/mck-sidebox.html
```diff
@@ -1,953 +1,2267 @@
...
-									</div>
-								</div>
-								
-							</div>
-						</div>
-						<div id="mck-tab-conversation" class="mck-row">
-							<h4 id="mck-conversation-title" class="mck-box-title mck-truncate">Conversations</h4>
-						</div>
-					</div>
+<div
```
devashishmamgain, line 1: can that container changes in the html and other javascript break the UI customization that customers might have done at their end?

devashishmamgain (changes requested):

devashishmamgain (approved):

PR #1322, opened 2025-06-18
Changed default to make socket connection only click on Chat icon

thakaresakshi121:
-  if user did not clicked on the chat widgeticon then  a disconnecting socket connection after one minute and reconnect once user clicks on chat icon.
-  for the new users  the connection only established after clickning on the chat widget

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 

### DIFF
--- a/webplugin/js/app/kommunicateCommons.js
+++ b/webplugin/js/app/kommunicateCommons.js
@@ -11,8 +11,7 @@ function KommunicateCommons() {
         CUSTOMER_CREATED_AT = optns.customerCreatedAt;
         USE_BRANDING = typeof optns.useBranding == 'boolean' ? optns.useBranding : true;
         WIDGET_SETTINGS = optns.widgetSettings;
-        KommunicateCommons.CONNECT_SOCKET_ON_WIDGET_CLICK =
-            optns.connectSocketOnWidgetClick || false;
+        KommunicateCommons.CONNECT_SOCKET_ON_WIDGET_CLICK = true;
     };
 
     _this.isTrialPlan = function (pricingPackage) {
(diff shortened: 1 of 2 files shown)


webplugin/js/app/kommunicateCommons.js
```diff
@@ -190,7 +189,7 @@ function KommunicateCommons() {
     _this.setWidgetStateOpen = function (isWidgetOpen) {
         if (KommunicateCommons.IS_WIDGET_OPEN === isWidgetOpen) return; // return if same value is already assigned to IS_WIDGET_OPEN.
         KommunicateCommons.IS_WIDGET_OPEN = isWidgetOpen;
-        if (IS_SOCKET_CONNECTED) {
+        if (IS_SOCKET_CONNECTED && isWidgetOpen) {
```
devashishmamgain, line 192: recheck this condition, it is for disconnecting

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -615,20 +615,25 @@ const firstVisibleMsg = {
...
                     );
                 }, this.SOCKET_DISCONNECT_TIMER_VALUE);
             },
             stop: function () {
+                console.debug('disconnection stopped');
                 clearTimeout(this.SOCKET_DISCONNECT_TIMEOUT);
             },
         };
-        var CONNECT_SOCKET_ON_WIDGET_CLICK = appOptions.connectSocketOnWidgetClick;
+        var CONNECT_SOCKET_ON_WIDGET_CLICK = true;
```
devashishmamgain, line 636: better to set appOptions.connectSocketOnWidgetClick = true on top

webplugin/js/app/kommunicateCommons.js
thakaresakshi121: this condition is for  stopping disconnection timeout.
if disconnection  setTimOut of 1 min started  and before one minute if someone clicked on chat widget then clearing the timeout

PR #1325, opened 2025-06-20
Fix Close icon/button location not changed when popupWidget value is changed

jaitens:
- Fix Close icon/button location not changed when popupWidget value is changed

### PR Checklist

### How was the code tested?

- locally in mobile screen and large screen

-

- when we pass popupwidget: true then close icon should appear in the launcher it self and close icon from top right corner should not be visible and it was happening because checks were missing to check popupwidget to conditionally apply the same

### Screenshot
- 
image
image

### DIFF
--- a/webplugin/js/app/kommunicate.js
+++ b/webplugin/js/app/kommunicate.js
@@ -630,8 +630,9 @@ $applozic.extend(true, Kommunicate, {
         Kommunicate.updateChatContext(chatContext);
     },
     setDefaultIframeConfigForOpenChat: function (isPopupEnabled) {
-        !kommunicateCommons.checkIfDeviceIsHandheld() &&
+        if (!kommunicateCommons.checkIfDeviceIsHandheld() && isPopupEnabled) {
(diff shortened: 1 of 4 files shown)


panwaranuj01 (approved):

webplugin/js/app/kommunicate.js
```diff
@@ -630,8 +630,9 @@ $applozic.extend(true, Kommunicate, {
         Kommunicate.updateChatContext(chatContext);
     },
     setDefaultIframeConfigForOpenChat: function (isPopupEnabled) {
-        !kommunicateCommons.checkIfDeviceIsHandheld() &&
+        if (!kommunicateCommons.checkIfDeviceIsHandheld() && isPopupEnabled) {
```
panwaranuj01, line 633: Screenshot 2025-06-23 at 6 16 26 PM

@jaitens why this condition is required here widget is breaking

panwaranuj01 (changes requested):

jaitens:
closing it not required anymore

PR #1331, opened 2025-07-21
fix source not coming in new line

jaitens:
- fix source not coming in new line

### PR Checklist

### How was the code tested?

-

-

- According to the Markdown specification, a single newline (\n) is treated as a space or a soft line break. It does not create a new paragraph. So, marked.js saw 'Hello\nWorld' and rendered it as if it were on the same line: <p>Hello World</p>.

### Screenshot
- 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved markdown formatting by adding extra spacing between lines in the output.


sakshamsinghalkommunicate:
**Tested now source are coming in new line** 
Screenshot 2025-07-24 at 12 06 24 PM

PR #1339, opened 2025-07-24
 fixed issue with socket reconnection after internet disconnected for longer time 

thakaresakshi121:

  cause : Calling ALSocket.reconnect() immediately on the online event doesn't work because the Applozic SDK hasn't fully restored its internal state (e.g. user session, socket) yet.

fix: Adding a short setTimeout gives the SDK enough time to reinitialize from localStorage, so reconnect() works reliably.

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -2367,7 +2367,10 @@ const firstVisibleMsg = {
                         'n-vis',
                         'vis'
                     );
-                    window.Applozic.ALSocket.reconnect();
+                    setTimeout(() => {
+                        console.log('reaconnecting socket');
+                        window.Applozic.ALSocket.reconnect();
+                    }, 500);
                 });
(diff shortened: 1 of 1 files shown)


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -2367,7 +2367,10 @@ const firstVisibleMsg = {
                         'n-vis',
                         'vis'
                     );
-                    window.Applozic.ALSocket.reconnect();
+                    setTimeout(() => {
+                        console.log('reaconnecting socket');
```
devashishmamgain, line 2371: what is the guarantee that this will work? and things will happen at within time frame?
can this be event based? wait for the initialization then call this, or keep doing retry?

devashishmamgain (changes requested):

devashishmamgain (approved):

PR #1361, opened 2025-07-31
fix

sakshamsinghalkommunicate:
-Sometime feedback comes lest sometime right
-Fix inconsistent positioning of the feedback icons in the chat UI to ensure they always appear at the bottom-right of the message container

### PR Checklist

### How was the code tested?

-tested by deplotion on test
Screenshot 2025-07-31 at 4 54 41 PM

-

-

### Screenshot
- 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the positioning of the feedback element to align more accurately with the message box, removing the previous minimum width constraint for better visual consistency.

### DIFF
--- a/webplugin/js/app/components/answer-feedback-service.js
+++ b/webplugin/js/app/components/answer-feedback-service.js
@@ -170,7 +170,7 @@ class AnswerFeedback {
         const msgBoxWidth = msgBox.offsetWidth;
 
         const feedbackWidth = msgBoxWidth + 40;
-        feedbackElement.style.left = `${feedbackWidth > 120 ? feedbackWidth : 120}px`;
+        feedbackElement.style.left = `${feedbackWidth}px`;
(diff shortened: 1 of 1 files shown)


webplugin/js/app/components/answer-feedback-service.js
```diff
@@ -170,7 +170,7 @@ class AnswerFeedback {
         const msgBoxWidth = msgBox.offsetWidth;
 
         const feedbackWidth = msgBoxWidth + 40;
-        feedbackElement.style.left = `${feedbackWidth > 120 ? feedbackWidth : 120}px`;
```
panwaranuj01, line 173: @thakaresakshi121 we have added this check for the agentic first right?
If yes, @sakshamsinghalkommunicate verify in the agenticfirst test phase and share the ss

PR #1388, opened 2025-09-18
Fix the multiple socket connection issue when opening the widget immediately

panwaranuj01:
- Fix the multiple socket connection issue when opening the widget immediately
- Added the test cases to test the socket connection issues

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 

## Summary by CodeRabbit

* **Bug Fixes**
  * Safer socket lifecycle: prevents overlapping inits, adds explicit initialized state, improves disconnect handling, and reduces reconnect/race failures.

* **Tests**
  * New end-to-end Playwright tests validating widget socket lifecycle: single-connection behavior, reconnect handling after manual disconnect, and proper cleanup.

* **Chores**
  * Updated bundled chat library to a newer compatible version.

### DIFF
--- a/playwright_test/tests/socketConnection.spec.js
+++ b/playwright_test/tests/socketConnection.spec.js
@@ -5,7 +5,7 @@ import { SCRIPT } from '../utils/kmScript';
 
 const isWidgetSocketUrl = (url) => /\/stomp/i.test(url) || /kommunicate\.(?:io|net)\/ws/i.test(url);
 
-const createSocketTracker = (page) => {
+const createSocketTracker = (page, log) => {
     const socketRecords = [];
 
     page.on('websocket', (ws) => {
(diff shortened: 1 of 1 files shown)


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -754,6 +755,10 @@ const firstVisibleMsg = {
                 mckInitializeChannel.onTypingStatus(resp);
             },
             onFeedbackClick: function (resp) {},
+            onDisconnect: function () {
+                IS_SOCKET_CONNECTED = false;
```
devashishmamgain, line 759: where are these getting used?
and what is the difference between connected and initialized?

devashishmamgain (approved):

PR #1443, opened 2026-01-14
TDH-3233 Away Message is not Coming in Widget

AashishBhatt5:
- 
image

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 

## Summary by CodeRabbit

* **Refactor**
  * Improved individual conversation detection for more reliable away-message behavior.
* **New Features**
  * Added versioned release outputs and exposed plugin version data for consumers.
* **Chores**
  * Streamlined build/deploy pipeline with optional CDN sync and cache-header steps; updated hosting cache rules.
* **Documentation**
  * Added a commented phone-field example in the demo configuration.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/webplugin/js/app/kommunicate-ui.js
+++ b/webplugin/js/app/kommunicate-ui.js
@@ -19,6 +19,15 @@ function getFaqClearButton() {
 function getBottomTabsManager() {
     return bottomTabsManagerRef;
 }
+function isConversationWindowNotActive() {
+    var sideboxContent = document.getElementById('mck-sidebox-content');
+    return (
+        !sideboxContent ||
+        !sideboxContent.classList ||
+        !sideboxContent.classList.contains('active-tab-conversations') ||
+        sideboxContent.classList.contains('active-subsection-conversation-list')
+    );
+}
 function setActiveSubsectionState(subsection) {
     var sideboxContent =
         typeof document !== 'undefined' &&
@@ -112,7 +121,7 @@ KommunicateUI = {
         }
     },
     populateAwayMessage: function (err, message) {
-        var conversationWindowNotActive = $applozic('#mck-tab-individual').hasClass('n-vis');
+        var conversationWindowNotActive = isConversationWindowNotActive();
         var closedConversation = $applozic('#mck-conversation-status-box').hasClass('vis');
         if (
             !err &&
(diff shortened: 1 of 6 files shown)


webplugin/js/app/kommunicate-ui.js
```diff
@@ -226,7 +235,7 @@ KommunicateUI = {
         });
     },
     showAwayMessage: function () {
-        var conversationWindowNotActive = $applozic('#mck-tab-individual').hasClass('n-vis');
+        var conversationWindowNotActive = isConversationWindowNotActive();
```
devashishmamgain: function name looks wrong, it looks like this is for case when individual conversation is active and away message will be shown.
verify and name function accordingly.

devashishmamgain (changes requested):
@AashishBhatt5 tested in both modern and classic layout?

webplugin/js/app/kommunicate-ui.js
AashishBhatt5: isConversationWindowNotActive() returns true when the chat window should be treated as inactive — meaning we’re either not on the conversations tab or we’re on the conversation list view (not inside an individual chat). It returns false only when an individual conversation is active.

AashishBhatt5:
Yes, Tested in both

webplugin/js/app/kommunicate-ui.js
devashishmamgain: no, this is checking following
sideboxContent.classList.contains('active-subsection-conversation-list')

which means when the Individual conversation thread is active and showing up.

devashishmamgain (changes requested):

AashishBhatt5:
@devashishmamgain 
active-subsection-conversation-list means the list view is active, not the individual thread. When the individual conversation is active, the class is active-subsection-conversation-individual.

Isn't this correct

devashishmamgain (dismissed):

devashishmamgain (dismissed):

webplugin/js/app/kommunicate-ui.js
```diff
@@ -112,13 +121,13 @@ KommunicateUI = {
         }
     },
     populateAwayMessage: function (err, message) {
-        var conversationWindowNotActive = $applozic('#mck-tab-individual').hasClass('n-vis');
+        var isIndividualConversation = isIndividualConversationActive();
         var closedConversation = $applozic('#mck-conversation-status-box').hasClass('vis');
```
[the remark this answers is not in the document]
devashishmamgain: this is the expected behaviour

AashishBhatt5:
https://github.com/Kommunicate-io/Kommunicate-Web-SDK/pull/1448

PR #1464, opened 2026-02-13
Fix faq visibility issue in chat widget

ayush-kommunicate:
#### The FAQ tab should be displayed in the chat widget only when there is at least 1 published FAQ article.
- Currently, the FAQ tab is visible even when:
  - The category is empty
  - The category contains only draft articles
- Expected behaviour: The FAQ tab should not be shown unless there is at least 1 published article.

#### When CSAT is disabled, the three-dot icon should not be visible.
- Currently, when CSAT is disabled, the three-dot icon is still visible and a blank div is rendered.
image

- Expected behaviour:
  - When CSAT is disabled, the icon should not be displayed in the modern layout.
  - In the classic layout, if both CSAT is disabled and the FAQ is empty, the icon should not be displayed.

### PR Checklist

### How was the code tested?

- Tested with the following scenarios for FAQ visibility:
  - 0 categories created
  - Empty category (no articles)
  - Category with all draft articles
  - Category with some draft and some published articles
  - Category with at least 1 published article
- Verified that the FAQ tab is shown only when at least 1 published article exists.

- Tested with the following scenarios for CSAT option visibility:
  - Disabled CSAT and verified in modern layout
  - Disabled CSAT and verified in classic layout with empty FAQ

- FAQ tab visibility was previously dependent on category existence instead of published article count.
- The three-dot container hide CSS was not working because the required CSS was missing/overridden.

- FAQ tab was getting displayed even when there were no published FAQ articles because it was only checking category presence.
- Missing CSS was causing the three-dot container to remain visible even when CSAT was disabled.

### Screenshot
- 

## Summary by CodeRabbit

* **Bug Fixes**
  * FAQ button now hides automatically when the knowledge base is empty or when no categories contain articles.

* **UI Changes**
  * Widget options visibility relaxed so options are no longer forcibly hidden in inactive tabs and can display in more tab/subsection states.

### DIFF
--- a/webplugin/knowledgebase/kb.js
+++ b/webplugin/knowledgebase/kb.js
@@ -41,15 +41,16 @@
                     success: function (data) {
                         response.status = 'success';
                         response.data = data.data || [];
-                        var hasArticles = response.data.some(function (category) {
-                            return category && Number(category.articleCount) > 0;
-                        });
-                        if (!hasArticles) {
-                            hideFAQBtn();
-                        }
 
                         if (options.success) {
-                            response.data.length === 0 && hideFAQBtn();
+                            var shouldHideFAQ =
+                                response.data.length === 0 ||
+                                !response.data.some(function (category) {
+                                    return category && Number(category.articleCount) > 0;
+                                });
+                            if (shouldHideFAQ) {
+                                hideFAQBtn();
+                            }
                             options.success(response);
                         }
                         return;
--- a/webplugin/knowledgebase/kb.js
+++ b/webplugin/knowledgebase/kb.js
@@ -41,6 +41,13 @@
                     success: function (data) {
                         response.status = 'success';
                         response.data = data.data || [];
+                        var hasArticles = response.data.some(function (category) {
+                            return category && Number(category.articleCount) > 0;
+                        });
+                        if (!hasArticles) {
+                            hideFAQBtn();
+                        }
+
                         if (options.success) {
                             response.data.length === 0 && hideFAQBtn();
                             options.success(response);
@@ -60,6 +67,7 @@
         };
 
         KommunicateKB.getArticles = function (options) {
+            console.log('getArticles called with categoryName: ' + options.data.categoryName);
             try {
                 var articles = [];
                 KommunicateKB.getFaqs({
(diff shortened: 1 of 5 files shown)


webplugin/knowledgebase/kb.js
```diff
@@ -41,6 +41,13 @@
                     success: function (data) {
                         response.status = 'success';
                         response.data = data.data || [];
+                        var hasArticles = response.data.some(function (category) {
```
devashishmamgain: this should move inside options.success and consolidate this with response.data.length === 0 check which is just below it. basically all conditions logic to decide whether to show faq or not should be at the same place just once.

devashishmamgain (changes requested):

devashishmamgain (dismissed):

webplugin/scss/components/_km-tab-visibility-conversations.scss
```diff
@@ -72,6 +72,9 @@
     #km-widget-options {
         display: flex !important;
     }
+    #km-widget-options.n-vis {
```
devashishmamgain: having n-vis class itself means display: none

figure out why it is displayed even after having n-vis
probably some code is setting n-vis and display block or vis too

devashishmamgain (changes requested):

webplugin/scss/components/_km-tab-visibility-conversations.scss
ayush-kommunicate: Sir, In the conversation-individual state, visibility is getting enforced by a highly specific selector:

#mck-sidebox-content.active-tab-conversations.active-subsection-conversation-individual #km-widget-options { display: flex !important; } (line no. 72)

Although .n-vis also applies display: none !important, the state-specific selector taking higher specificity. As a result, it can take precedence in the CSS and keep the 3-dot options trigger visible despite the n-vis class.

devashishmamgain: @ayush-kommunicate check if adding !important is required on display: flex !important; ? 
try removing that

ayush-kommunicate: Yes sir without !important it remain hidden even when CSAT is enable.

devashishmamgain (approved):

PR #1469, opened 2026-02-19
TDH 3221 Chat widget duplicate message on slow internet

ayush-kommunicate:
1. The chat widget should not show duplicate messages on slow internet.
2. Added an option in `demo2.html` to set the widget position (left/right), which helps in testing easily.
3. The popup message (showed on new message) close button should be properly aligned with the popup message.

### Solution 
For chat widget duplicate message, it first check weaterh response has came from ther http req. then update the mesage if not it wait for 4 sec and check every second and once it get the response of http then only it update the echo message. 

### PR Checklist

#### 1. Duplicate message issue
- Set network to 2G/3G (slow internet).
- Sent a message from the widget.
- Verified that the message does not appear twice in the UI.

#### 2. Widget position option
- Opened `demo2.html`.
- Changed the widget position to left and right.
- Verified that the widget opens correctly at the selected position.

#### 3. Popup close button alignment
- Opened the widget and triggered a popup message.
- Verified that the close button is properly aligned with the popup message.

- On slow internet, the socket echo may arrive before the HTTP response, which can cause message duplication if optimistic messages are not properly mapped.

#### 1. Duplicate message issue (before fix):
- User sends a message → UI renders an optimistic message bubble with a random key.
- HTTP response is slow.
- Socket echo arrives first with the server-generated key.
- The echo handler does not find the optimistic bubble (keys differ), so it adds a second bubble.
- Later, the HTTP response updates the original optimistic bubble key to the server key, but the duplicate bubble remains.

#### Solution:  
When a user sends a message, the UI first renders an optimistic message bubble with a temporary key. If the HTTP response is slow and the socket echo arrives first with the server-generated key, the system now first checks whether the HTTP response has been received. If the response has already arrived, it updates the existing optimistic message. If not, it waits for up to 4 seconds and checks every second for the HTTP response. Once the HTTP response is received, only then does it update the original message and its status instead of adding a new bubble immediately . This prevents duplicate messages from appearing in the UI.
#### 2. Popup close button alignment issue:
- In the chat widget, when a new message popup appears, the close button was not properly aligned with the popup message due to CSS styling issues.

### Screenshot
- 
image

image

Widget position feature should be like:
image

NOTE: Make sure you're comparing your branch with the correct base branch.

## Summary by CodeRabbit

* **New Features**
  * Added widget position setting (Left/Right) that persists across launches and reflects in previews.
  * More resilient message status handling: improved delivered/read indicators, retries, and sync of unread counts and conversation presence.
  * Better handling and clearer feedback for encrypted or problematic file attachments.

* **Style**
  * Adjusted chat close button size and positioning for improved alignment and offset.


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -12027,6 +12030,24 @@ const firstVisibleMsg = {
                             }
                         }
                     } else if (messageType === 'APPLOZIC_02' && !(message.contentType == 102)) {
+                        var clientMessageId =
+                            message.metadata && message.metadata.KM_CLIENT_MESSAGE_ID;
+                        if (clientMessageId && $applozic('.' + clientMessageId).length) {
+                            var $clientMsg = $applozic('.' + clientMessageId).first();
+                            if (message.conversationId) {
+                                $mck_msg_inner.data('mck-conversationid', message.conversationId);
```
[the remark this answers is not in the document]
ayush-kommunicate: This is a valid variable defined in line no. 7813 (var $mck_msg_inner = $applozic('#mck-message-cell .mck-message-inner') inside same function;

devashishmamgain:
@ayush-kommunicate can you add details on the logic used to avoid that

@devashishmamgain  added in description sir.

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -12027,6 +12030,24 @@ const firstVisibleMsg = {
                             }
                         }
                     } else if (messageType === 'APPLOZIC_02' && !(message.contentType == 102)) {
+                        var clientMessageId =
+                            message.metadata && message.metadata.KM_CLIENT_MESSAGE_ID;
+                        if (clientMessageId && $applozic('.' + clientMessageId).length) {
+                            var $clientMsg = $applozic('.' + clientMessageId).first();
```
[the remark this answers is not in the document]
ayush-kommunicate: Resolved.

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -14853,25 +14914,37 @@ const firstVisibleMsg = {
...
+                    if ($readStatus.length) {
+                        $readStatus
+                            .removeClass('mck-pending-icon')
+                            .removeClass('mck-sent-icon')
+                            .removeClass('mck-delivered-icon')
+                            .addClass('mck-read-icon');
+                        mckMessageLayout.addTooltip(readKey);
+                    } else {
+                        pendingMessageStatusUpdates[readKey] = 'read';
+                    }
```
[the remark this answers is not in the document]
ayush-kommunicate: Resolved.

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -12118,6 +12115,70 @@ const firstVisibleMsg = {
...
+            var applyPendingStatusUpdate = function (messageKey) {
+                if (!messageKey || !pendingMessageStatusUpdates[messageKey]) {
+                    return;
+                }
+                var statusType = pendingMessageStatusUpdates[messageKey];
+                var $status = $applozic('.' + messageKey + ' .mck-message-status');
+                if (!$status.length) {
+                    return;
+                }
+                if (statusType === 'delivered') {
```
devashishmamgain: similar code is there in other place too to add and remove these pending, delivered, read icon, avoid duplicate code. make it common reusable code.

devashishmamgain (changes requested):

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -14916,14 +14930,13 @@ const firstVisibleMsg = {
...
                     var $deliveredStatus = $applozic('.' + deliveredKey + ' .mck-message-status');
-                    if ($deliveredStatus.length) {
-                        $deliveredStatus
-                            .removeClass('mck-pending-icon')
-                            .removeClass('mck-sent-icon')
-                            .addClass('mck-delivered-icon')
-                            .attr('title', 'delivered');
-                        mckMessageLayout.addTooltip(deliveredKey);
-                    } else {
+                    if (
```
devashishmamgain, line 15149: is this change correct?

devashishmamgain (changes requested):

devashishmamgain:
@ayush-kommunicate is this change correct?

Screenshot 2026-02-27 at 9 30 19 PM

@devashishmamgain yes sir.

webplugin/js/app/mck-sidebox-1.0.js
ayush-kommunicate: Yes sir function will retun false if status.length is 0.

devashishmamgain: do we need this if clause?
pendingMessageStatusUpdates[...] can be moved to that same function?

devashishmamgain (changes requested):

devashishmamgain (approved):

PR #1478, opened 2026-03-03
TDH-3462: Fix chatwidget error

adityaKommunicate:
-Fix ReferenceError: loadChat is not defined that crashes the widget during preview re‑init (layout/launcher changes) by ensuring the widget bundle calls a globally defined loadChat, and verify locally against the patched SDK bundle.

### PR Checklist

### How was the code tested?

-Ran dashboard locally.
Switched between Classic ↔ Modern layout and changed launcher while logged in.
Confirmed widget no longer disappears and no loadChat is not defined error.
Verified widget script loaded from http://localhost:3030/kommunicate-widget-3.0.min.js via DevTools Network.

-The dashboard does not call loadChat directly; the SDK calls it during pre‑lead re‑initialization, so the error is rooted in SDK scope, not dashboard config.

-Root cause: in the SDK bundle, loadChat was referenced during pre‑lead re‑init from a different closure where it wasn’t in scope, causing ReferenceError on re‑initialization events. The fix is to expose loadChat on window and call window.loadChat.

### Screenshot
- NA

## Summary by CodeRabbit

* **Refactor**
  * Made the chat widget triggerable from other scripts so it can be launched or relaunched externally.
  * Ensured consistent reinitialization so show/hide and widget state are reliably reset when reopened.
  * Preserved existing chat behavior while exposing a stable external entry point for launching the widget.

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -4127,7 +4127,7 @@ const firstVisibleMsg = {
                 }
                 console.log('[PRE-LEAD] loadChat completed, widget re-launched');
             }
-            if (typeof window !== 'undefined' && !window.loadChat) {
+            if (typeof window !== 'undefined' && typeof window.loadChat !== 'function') {
                 window.loadChat = loadChat;
             }
             /*  To trigger welcome event of a bot.
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -2306,7 +2306,7 @@ const firstVisibleMsg = {
                                 chatNotificationMailSent: true,
                             };
                             PRE_CHAT_LEAD_COLLECTION_POPUP_ON = false;
-                            mckInit.initialize(options, window.loadChat);
+                            mckInit.initialize(options, mckInit.loadChatCallback);
                             return false;
                         }
                         var kmAnonymousChatLauncher = document.getElementById(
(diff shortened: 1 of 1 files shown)


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -4131,6 +4127,9 @@ const firstVisibleMsg = {
                 }
                 console.log('[PRE-LEAD] loadChat completed, widget re-launched');
             }
+            if (typeof window !== 'undefined' && !window.loadChat) {
+                window.loadChat = loadChat;
```
devashishmamgain: not sure if this is the right solution, are there other places in chat widget code where "window." is used before function calls?
please recheck why window. is required here for this use case, other functions seems to be working without that in this repo

adityaKommunicate: window.loadChat guarantees the callback is globally reachable, regardless of how the bundle rearranges scopes. Most other functions work without window. because they’re accessed via:
Instance methods (_this.someFn)
Namespaced objects (Kommunicate, mckMessageService, etc.)
Or they remain in the same closure after bundling.
loadChat is a free identifier passed as a callback, which is exactly the pattern that breaks when the bundler splits closures.if (typeof window !== 'undefined' && typeof window.loadChat !== 'function') {
  window.loadChat = loadChat;
}
this part is a guard so we only define it if missing or invalid and other callbacks are passed by reference within the same scope

devashishmamgain: _loadChat is a free identifier passed as a callback_
>> this is passed externally or is passed by this codebase itself?

devashishmamgain (approved):

PR #1486, opened 2026-03-11
TDH-3284-Remove unused data from metadata

ayush-kommunicate:
- In the Group Create API, These metadata fields are sent with empty values in every request. For support groups, this metadata is not used anywhere. The backend currently removes this data before storing it in the database but still returns it in the response.

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 

## Summary by CodeRabbit

* **Refactor**
  * Cleaned up unused metadata field initializations in conversation creation to streamline the codebase.


devashishmamgain (approved):
@ayush-kommunicate so these are anyways not getting saved in the database?

ayush-kommunicate:
> @ayush-kommunicate so these are anyways not getting saved in the database?

Yes. These are not getting saved in database.

PR #1491, opened 2026-03-25
customization chat issue

adityaKommunicate:
-Ensure the chat widget preview (Customization & Playground) loads correctly even when domain restrictions are enabled.
-Fix chat widget preview loading in dashboard customization and local demo while keeping domain restriction enforcement safe for srcdoc/about:blank embeds.

### PR Checklist

### How was the code tested?

-Dashboard customization preview on http://127.0.0.1:3000/settings/chat-widget-customization.
Local preview via example/demo2.html on http://127.0.0.1:3030/example/demo2.html.
Verified domain restriction logic with/without allowedDomains, including srcdoc/about:blank iframe contexts.

-Safari (and some older browsers) use about:blank when srcdoc is unsupported; preview host resolution must handle both about:srcdoc and about:blank.

-The widget was not loading in customization preview because domain checks were using the iframe’s about:srcdoc/about:blank context, which doesn’t provide a real hostname. This caused allowedDomains evaluation to fail or be bypassed. Resolving the effective hostname from the nearest non‑srcdoc/about:blank parent fixes it while still enforcing domain restrictions.
- Root cause: Customization preview loads inside `about:srcdoc`, so `window.location.hostname` is empty. The SDK domain restriction logic used this value, causing the widget to be blocked when allowlist was enabled.

### Screenshot
- NA

## Summary by CodeRabbit

* **Bug Fixes**
  * Accept preview on localhost/127.0.0.1 by initializing and appending the preview host to allowed domains when missing.
  * Improve embedded preview handling by skipping domain-restriction checks for srcdoc/about:blank contexts and using the first accessible parent frame location when available.
  * Expand recognized service domains to include dashboard.kommunicate.io and dashboard-test.kommunicate.io.

### DIFF
--- a/example/demo2.html
+++ b/example/demo2.html
@@ -771,20 +771,6 @@
                 ) {
                     globalSettings.widgetSettings = {};
                 }
-                // for demo when domain restriction is enabled
-                const previewHost = window.location.hostname;
-                if (previewHost === 'localhost' || previewHost === '127.0.0.1') {
-                    const existingAllowed = Array.isArray(
-                        globalSettings.widgetSettings.allowedDomains
-                    )
-                        ? globalSettings.widgetSettings.allowedDomains
-                        : [];
-                    if (!existingAllowed.includes(previewHost)) {
-                        globalSettings.widgetSettings.allowedDomains = existingAllowed.concat(
-                            previewHost
-                        );
-                    }
-                }
                 globalSettings.widgetSettings.position = config.widgetPosition;
 
                 if (config.voiceModeEnabled) {
--- a/example/demo2.html
+++ b/example/demo2.html
@@ -771,6 +771,20 @@
                 ) {
                     globalSettings.widgetSettings = {};
                 }
+                // for demo when domain restriction is enabled
+                const previewHost = window.location.hostname;
+                if (previewHost === 'localhost' || previewHost === '127.0.0.1') {
+                    const existingAllowed = Array.isArray(
+                        globalSettings.widgetSettings.allowedDomains
+                    )
+                        ? globalSettings.widgetSettings.allowedDomains
+                        : [];
+                    if (!existingAllowed.includes(previewHost)) {
+                        globalSettings.widgetSettings.allowedDomains = existingAllowed.concat(
+                            previewHost
+                        );
+                    }
+                }
                 globalSettings.widgetSettings.position = config.widgetPosition;
 
                 if (config.voiceModeEnabled) {
--- a/webplugin/js/app/km-utils.js
+++ b/webplugin/js/app/km-utils.js
@@ -219,9 +219,8 @@ KommunicateConstants = {
     },
     KOMMUNICATE_DOMAINS: [
         'kommunicate.io',
+        'dashboard.kommunicate.io',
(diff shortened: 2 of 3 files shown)


webplugin/js/app/km-utils.js
```diff
@@ -217,7 +217,12 @@ KommunicateConstants = {
         LEFT: 'left',
         RIGHT: 'right',
     },
-    KOMMUNICATE_DOMAINS: ['kommunicate.io'],
+    KOMMUNICATE_DOMAINS: [
+        'kommunicate.io',
+        'dashboard-test.kommunicate.io',
+        'localhost',
+        '127.0.0.1',
+    ],
```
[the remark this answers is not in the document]
adityaKommunicate: okay

webplugin/js/app/km-utils.js
```diff
@@ -217,7 +217,14 @@ KommunicateConstants = {
         LEFT: 'left',
         RIGHT: 'right',
     },
-    KOMMUNICATE_DOMAINS: ['kommunicate.io'],
+    KOMMUNICATE_DOMAINS: [
+        'kommunicate.io',
```
devashishmamgain, line 221: is it possible to have *.kommunicate.io so that all sub domains work?

devashishmamgain (changes requested):

webplugin/js/app/km-utils.js
adityaKommunicate: kommunicate.io is enough, added as safety measure

adityaKommunicate: removed

example/demo2.html
```diff
@@ -855,6 +771,20 @@ <h2>Live preview area</h2>
                 ) {
                     globalSettings.widgetSettings = {};
                 }
+                // for demo when domain restriction is enabled
+                const previewHost = window.location.hostname;
```
devashishmamgain: are we explicitly allowing localhost 127.0.0.1 here?
with this approach then anyone will be able to use others chat widget. remove that

webplugin/js/app/mck-app.js
```diff
@@ -424,7 +424,26 @@ function ApplozicSidebox() {
                     : widgetSettings.disableChatWidget; // Give priority to appOptions over API data.
 
             var allowedDomains = widgetSettings.allowedDomains;
-            var hostname = parent.window.location.hostname.toLowerCase();
+            var contextWindow = window;
+            var hostname = '';
+            try {
+                while (
```
devashishmamgain, line 430: not able to understand this, can you explain what you are trying here? why do we need while? any chance that this can go in an infinite loop?

devashishmamgain (changes requested):

webplugin/js/app/mck-app.js
adityaKommunicate: So sir we were getting issue in Customization Preview in dashboard which has iframe having url about:srcdoc which doent have a real hostname so allowed‑domains check blocked it. The fix is to find the  parent window iframe which has a real url thats why using while loop and no infinite loop would not happen checked with sentry also.

devashishmamgain (approved):

PR #1494, opened 2026-03-26
Fix-msz-delivery-icon

ayush-kommunicate:
- The chatbot message delivery icon is not displaying correctly.
- Currently, the same icon is shown for all statuses (pending, sent, delivered, read):

image

- It should display different icons for each status as expected:

image

### PR Checklist

### How was the code tested?
- Tested locally by verifying message delivery states (pending, sent, delivered, read) and ensuring correct icons are displayed.

- Learned that using SVG `<use>` with symbols does not preserve inner `<path>` structure for styling via class selectors.

- The issue occurred after moving the icon to `mck-icons.js` as `#icon-71`.
- The icon was rendered using `<use>`, which does not retain the internal `<path>` elements in a way that allows CSS class-based styling.
- Since delivery states (pending, sent, delivered, read) depend on classes applied to individual `<path>` elements, all states ended up showing the same icon.
- The fix was to keep the SVG centralized in `mck-icons.js` but render the actual `<path>` markup directly in the message template instead of using the sprite reference.

### Screenshot
- (Add updated screenshot here if needed)

### Note
- Make sure you are comparing your branch with the correct base branch.

## Summary by CodeRabbit

* **Refactor**
  * Reorganized how message status icons are internally managed for improved code maintainability.


AashishBhatt5 (approved):

AashishBhatt5:
@ayush-kommunicate has this been changed in the latest commit?
No this was changed in some other pr changes.

PR #1505, opened 2026-04-02
Fix race condition in conversation create

ayush-kommunicate:
- When Kommunicate.startConversation() is used in the script, the conversation is not displayed in the chat widget and is causing a race condition issue.
https://intentive-ai.slack.com/archives/C0178DHF2RH/p1775069258897659

It should directly open a new conversation in the chat widget.

- On closing the widget, the logo size gets cut when the window is resized.

image

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved preservation of conversation state when switching between tabs
  * Enhanced detection and handling of closed chat widget interface states
  * Refined tab navigation behavior for more consistent user experience

### DIFF
--- a/webplugin/js/app/km-bottom-tabs.js
+++ b/webplugin/js/app/km-bottom-tabs.js
@@ -312,10 +312,21 @@
         function getActiveConversationId() {
             var messageInner =
                 documentRef && documentRef.querySelector('#mck-message-cell .mck-message-inner');
+            var messageInnerDataId = null;
+            var jq =
+                (typeof w !== 'undefined' && (w.$applozic || w.jQuery)) ||
+                (typeof $applozic !== 'undefined' && $applozic) ||
+                (typeof jQuery !== 'undefined' && jQuery);
+            if (messageInner && typeof jq === 'function') {
+                try {
+                    messageInnerDataId = jq(messageInner).data('mck-id');
+                } catch (e) {}
+            }
             return (
                 messageInner &&
                 (messageInner.getAttribute('data-mck-id') ||
-                    (messageInner.dataset && messageInner.dataset.mckId))
+                    (messageInner.dataset && messageInner.dataset.mckId) ||
+                    messageInnerDataId)
             );
         }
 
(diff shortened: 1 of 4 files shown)


thakaresakshi121 (dismissed):

thakaresakshi121 (dismissed):

webplugin/js/app/km-bottom-tabs.js
```diff
@@ -312,10 +312,21 @@
         function getActiveConversationId() {
             var messageInner =
                 documentRef && documentRef.querySelector('#mck-message-cell .mck-message-inner');
+            var messageInnerDataId = null;
+            var jq =
```
devashishmamgain: why are there so many undefined checks required for different things?
what is the use case when these w is coming undefined, in which case $applozic is coming undefined? and when is jQuery coming as undefined?
Also, check why jq=='function' check is required?

webplugin/js/app/km-bottom-tabs.js
```diff
@@ -312,10 +312,21 @@
...
+                (typeof w !== 'undefined' && (w.$applozic || w.jQuery)) ||
+                (typeof $applozic !== 'undefined' && $applozic) ||
+                (typeof jQuery !== 'undefined' && jQuery);
+            if (messageInner && typeof jq === 'function') {
+                try {
```
devashishmamgain: when will this code can potentionally throw any error?
we are fetching data attribute at many places, but havn't seen such check anywhere, if data is not there it is simplify return undefined or null, but wouldn't throw the error, as long as messageInner is present.

why is jquery usage required to get the data attribute? core javascript can also do that?

webplugin/js/app/km-bottom-tabs.js
```diff
@@ -453,10 +464,21 @@
                         var latestMessageInner =
                             documentRef &&
                             documentRef.querySelector('#mck-message-cell .mck-message-inner');
+                        var latestMessageInnerDataId = null;
+                        var jq =
+                            (typeof w !== 'undefined' && (w.$applozic || w.jQuery)) ||
```
devashishmamgain: same comments as above.

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -955,7 +955,17 @@ const firstVisibleMsg = {
                 }
                 // Voice socket topic lifecycle on connect (safe lookup, no scoped jquery dependency).
                 var messageInner = document.querySelector('#mck-message-cell .mck-message-inner');
-                var currentTabId = messageInner ? messageInner.getAttribute('data-mck-id') : null;
+                var currentTabId = null;
+                if (messageInner) {
+                    var messageInnerDataId =
```
devashishmamgain: same same thing
also duplicate code, copy pasted 3 times

devashishmamgain (changes requested):

devashishmamgain (approved):

PR #1506, opened 2026-04-06
TDH-3613 Remove Eleven labs code from chat widget

AashishBhatt5:
-

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 

## Summary by CodeRabbit

* **Refactor**
  * Unified voice/text handling into a single omnichannel transport, simplifying TTS/STT flows for more consistent behavior.
* **Bug Fixes**
  * Improved recognition fallbacks so native speech falls back to omnichannel when unavailable, reducing unreliable legacy paths.
  * Added explicit error reporting when TTS fails to produce audio, improving failure visibility.

### DIFF
--- a/webplugin/js/app/voice/mck-voice.js
+++ b/webplugin/js/app/voice/mck-voice.js
@@ -370,33 +370,27 @@ class MckVoice {
                 return;
             }
 
-            if (this.activeRecognitionMode === 'omnichannel') {
-                if (queueItem.ttsError && !queueItem.ttsBlob && !queueItem.ttsPromise) {
-                    const prefetchError = queueItem.ttsError;
-                    queueItem.ttsError = null;
-                    throw prefetchError;
-                }
-                if (!queueItem.ttsBlob && !queueItem.ttsPromise) {
-                    queueItem.ttsPromise = kmVoice
-                        .textToVoice(spokenText)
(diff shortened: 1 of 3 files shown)


webplugin/js/app/voice/mck-voice.js
```diff
@@ -370,33 +370,27 @@ class MckVoice {
...
-                        .catch((error) => {
-                            queueItem.ttsPromise = null;
-                            queueItem.ttsError = error;
-                            throw error;
-                        });
-                }
-                const wavBlob = queueItem.ttsBlob || (await queueItem.ttsPromise);
-                this.playAudioBlobWithQueue(wavBlob);
-                return;
+            if (queueItem.ttsError && !queueItem.ttsBlob && !queueItem.ttsPromise) {
```
devashishmamgain, line 373: omnichannel check will be required, we have native browser support as well

devashishmamgain (dismissed):

webplugin/js/app/voice/mck-voice.js
AashishBhatt5: The native path is already handled earlier via shouldUseNativeSpeechSynthesis()

kandpal025 (approved):

PR #1518, opened 2026-04-21
TDH-3659:Chat widget full view is not coming in iPhone, it going beyond the viewable space both horizontally and vertically

adityaKommunicate:
-Fix the mobile chat widget layout on iPhone so that when the keyboard opens, the widget stays aligned to the visible viewport and does not overflow off-screen.

### PR Checklist

### How was the code tested?

-Tested locally on iPhone

-On iPhone/mobile browsers, opening the keyboard changes the visual viewport dynamically, so relying only on static CSS sizing is not enough for the widget iframe to stay aligned.

-The widget iframe was not syncing with the mobile browser’s visible viewport when the keyboard opened, which caused incorrect iframe dimensions/positioning and made the widget overflow outside the screen on iPhone.

### Screenshot
- 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed iframe sizing and display on mobile/fullscreen by keeping its dimensions in sync with the visual viewport.
  * Prevented iframe removal issues by ensuring teardown stops viewport tracking when closed.

* **Improvements**
  * Improved compatibility with iOS/WebKit on mobile keyboards by dynamically updating iframe position and size during viewport changes.

### DIFF
--- a/server/config/config-env.js
+++ b/server/config/config-env.js
@@ -6,7 +6,7 @@ const configEnv = {
             kommunicateBaseUrl: 'https://api-test.kommunicate.io',
             omnichannelBaseUrl: 'https://omni-channel-test.kommunicate.io',
             botPlatformApi: 'https://bots-test.kommunicate.io',
-            hostUrl: 'http://10.10.24.82:3030',
+            hostUrl: 'http://localhost:3030',
             dashboardUrl: 'https://dashboard-test.kommunicate.io',
         },
         pluginProperties: {
--- a/server/config/config-env.js
+++ b/server/config/config-env.js
@@ -6,7 +6,7 @@ const configEnv = {
             kommunicateBaseUrl: 'https://api-test.kommunicate.io',
             omnichannelBaseUrl: 'https://omni-channel-test.kommunicate.io',
             botPlatformApi: 'https://bots-test.kommunicate.io',
-            hostUrl: 'http://localhost:3030',
+            hostUrl: 'http://10.10.24.82:3030',
             dashboardUrl: 'https://dashboard-test.kommunicate.io',
         },
         pluginProperties: {
--- a/webplugin/plugin.js
+++ b/webplugin/plugin.js
@@ -413,11 +413,11 @@ function attachMobileKeyboardViewportFix(iframeElement) {
         'max-height',
     ];
     var syncFrame = function () {
+        if (!mobileViewport.matches) return;
         if (
             !iframeElement.classList.contains('kommunicate-iframe-enable-media-query') ||
             iframeElement.classList.contains('km-iframe-closed') ||
-            iframeElement.getAttribute('data-km-widget-container') === 'true' ||
-            !mobileViewport.matches
+            iframeElement.getAttribute('data-km-widget-container') === 'true'
         ) {
             frameProperties.forEach(function (property) {
                 iframeElement.style.removeProperty(property);
--- a/webplugin/plugin.js
+++ b/webplugin/plugin.js
@@ -397,7 +397,6 @@ function attachMobileKeyboardViewportFix(iframeElement) {
     if (!window.visualViewport) {
         return;
     }
-
     var viewport = window.visualViewport;
     var mobileViewport = window.matchMedia('(max-width: 600px)');
     var frameProperties = [
@@ -410,7 +409,6 @@ function attachMobileKeyboardViewportFix(iframeElement) {
         'max-width',
         'max-height',
     ];
-
     var syncFrame = function () {
         if (
             !iframeElement.classList.contains('kommunicate-iframe-enable-media-query') ||
@@ -423,7 +421,6 @@ function attachMobileKeyboardViewportFix(iframeElement) {
             });
             return;
         }
-
         [
             ['top', viewport.offsetTop + 'px'],
             ['left', viewport.offsetLeft + 'px'],
@@ -449,7 +446,6 @@ function attachMobileKeyboardViewportFix(iframeElement) {
     });
     syncFrame();
 }
-
 // Create element iframe for kommunicate widget
 function createKommunicateIframe() {
     if (document.getElementById(kmCustomElements.iframe.id)) {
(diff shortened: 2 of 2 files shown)


webplugin/plugin.js
```diff
@@ -44,9 +44,7 @@ var kmCustomIframe =
     '   top: 0;' +
     '   left: 0px !important;' +
     '   border-radius: 0px;' +
-    '   height: 100% !important;' +
-    '   width: 100% !important;' +
-    '   max-height: 100% !important;' +
+    '   width: 100vw !important;' +
     '} } \n' +
     '.km-iframe-notification{ ' +
```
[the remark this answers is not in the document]
adityaKommunicate: not valid

webplugin/plugin.js
```diff
@@ -392,6 +393,67 @@ function isIOSWebKitBrowser(userAgent) {
...
+            ['width', viewport.width + 'px'],
+            ['height', viewport.height + 'px'],
+            ['max-width', viewport.width + 'px'],
+            ['max-height', viewport.height + 'px'],
+        ].forEach(function (entry) {
+            iframeElement.style.setProperty(entry[0], entry[1], 'important');
+        });
+    };
+
+    var onViewportChange = function () {
```
adityaKommunicate, line 450: when the viewport changes (resize / scroll)
don’t run syncFrame() immediately in the event callback
ask the browser to run it in the next animation frame

webplugin/plugin.js
```diff
@@ -392,6 +393,64 @@ function isIOSWebKitBrowser(userAgent) {
...
+        'right',
+        'bottom',
+        'width',
+        'height',
+        'max-width',
+        'max-height',
+    ];
+
+    var syncFrame = function () {
+        if (
```
adityaKommunicate, line 425: Checks whether the widget is not in the exact state where the iPhone keyboard fix should run.

These conditions mean:

mobile open class is missing
or widget is closed
or widget is embedded in container mode
or viewport is no longer mobile width

webplugin/plugin.js
```diff
@@ -392,6 +393,64 @@ function isIOSWebKitBrowser(userAgent) {
...
+            !mobileViewport.matches
+        ) {
+            iframeElement.classList.contains('km-iframe-closed') &&
+                frameProperties.forEach(function (property) {
+                    iframeElement.style.removeProperty(property);
+                });
+            return;
+        }
+
+        [
```
adityaKommunicate, line 436: Builds the inline styles needed for the current visible viewport and loops over them.

webplugin/plugin.js
```diff
@@ -392,6 +396,64 @@ function isIOSWebKitBrowser(userAgent) {
     return isIOSDevice(ua) && /AppleWebKit/i.test(ua);
 }
 
+function attachMobileKeyboardViewportFix(iframeElement) {
+    if (!window.visualViewport) {
```
adityaKommunicate: desktop issue

server/config/config-env.js
```diff
@@ -6,7 +6,7 @@ const configEnv = {
             kommunicateBaseUrl: 'https://api-test.kommunicate.io',
             omnichannelBaseUrl: 'https://omni-channel-test.kommunicate.io',
             botPlatformApi: 'https://bots-test.kommunicate.io',
-            hostUrl: 'http://localhost:3030',
+            hostUrl: 'http://10.10.24.82:3030',
```
devashishmamgain: ?

devashishmamgain: revert

devashishmamgain (changes requested):
@adityaKommunicate please check if these are valid comments
High: the new keyboard workaround is enabled for every browser that exposes window.visualViewport, not just iPhone/WebKit, because it is attached unconditionally in plugin.js (line 399) and called on every iframe creation in plugin.js (line 517). If the goal is only “iPhone message field focus pushes the top out of view”, this broadens the blast radius to Android/mobile Chrome and any other small-screen browser with visualViewport.

High: the CSS changes in plugin.js (line 41) and plugin.js (line 71) change the default mobile iframe layout for all sessions, not just when the iPhone keyboard is open. Switching to 100vw and 100dvh makes the open widget full-viewport on mobile even before the message field is focused, which is a broader product change than the stated bug fix.

Medium: the viewport-fix cleanup is incomplete when the viewport leaves the mobile breakpoint. In plugin.js (line 415), syncFrame() returns immediately on !mobileViewport.matches, so any previously applied inline top/left/width/height/max-* values remain in place after rotation or widening past 600px. That can leave the iframe pinned to stale dimensions instead of falling back to the normal desktop rules.

adityaKommunicate:
> @adityaKommunicate please check if these are valid comments High: the new keyboard workaround is enabled for every browser that exposes window.visualViewport, not just iPhone/WebKit, because it is attached unconditionally in plugin.js (line 399) and called on every iframe creation in plugin.js (line 517). If the goal is only “iPhone message field focus pushes the top out of view”, this broadens the blast radius to Android/mobile Chrome and any other small-screen browser with visualViewport.
> 
> High: the CSS changes in plugin.js (line 41) and plugin.js (line 71) change the default mobile iframe layout for all sessions, not just when the iPhone keyboard is open. Switching to 100vw and 100dvh makes the open widget full-viewport on mobile even before the message field is focused, which is a broader product change than the stated bug fix.
> 
> Medium: the viewport-fix cleanup is incomplete when the viewport leaves the mobile breakpoint. In plugin.js (line 415), syncFrame() returns immediately on !mobileViewport.matches, so any previously applied inline top/left/width/height/max-* values remain in place after rotation or widening past 600px. That can leave the iframe pinned to stale dimensions instead of falling back to the normal desktop rules.

tackled all three, comments were valid

webplugin/plugin.js
```diff
@@ -276,6 +276,9 @@ function removeKommunicateScripts() {
             commons.cleanupIframeResizeListener(kommunicateIframe);
         }
     }
+    kommunicateIframe &&
```
adityaKommunicate, line 279: checks if kommunicateIframe exists
check if a cleanup function named __kmDetachViewportFix was attached to it
if widget iframe is being removed it runs the cleanup for the mobile viewport fix

devashishmamgain (approved):

thakaresakshi121 (approved):

PR #1528, opened 2026-05-29
TDH-2552-Support-custom-widget-position

ayush-kommunicate:
- Users can set the widget position from widget customization, and the widget reads those values from app settings and applies the position accordingly.

Dashboard pr : https://github.com/Kommunicate-io/Kommunicate-Dashboard/pull/3884

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 

## Summary by CodeRabbit

* **New Features**
  * Sidebox widgets now support custom positioning with normalized pixel values and automatic left/right placement; respects when a widget iframe is not present.

* **Bug Fixes**
  * Consistent iframe height/layout adjustments across modes for improved display.
  * Removed forced minimum height in pre-lead flow for better sizing behavior.

### DIFF
--- a/webplugin/js/app/mck-app.js
+++ b/webplugin/js/app/mck-app.js
@@ -413,9 +413,7 @@ function ApplozicSidebox() {
     }
 
     function getWidgetCustomPositionProperties(widgetSettings) {
-        var horizontalPositionProperty =
-            widgetSettings.position === KommunicateConstants.POSITION.LEFT ? 'left' : 'right';
-        return ['bottom', horizontalPositionProperty];
+        return ['bottom', widgetSettings.position];
     }
 
     function applyWidgetCustomPosition(widgetSettings) {
--- a/webplugin/js/app/mck-app.js
+++ b/webplugin/js/app/mck-app.js
@@ -399,6 +399,58 @@ function ApplozicSidebox() {
         }
     }
 
+    var widgetPositionProperties = ['right', 'bottom', 'left'];
+
+    function normalizeWidgetPositionValue(value) {
+        return value == null || value === '' || !isFinite(value) ? null : Number(value) + 'px';
+    }
+
+    function resetWidgetCustomPosition(kommunicateIframe) {
+        widgetPositionProperties.forEach(function (property) {
+            kommunicateIframe.style.removeProperty(property);
+        });
+        kommunicateIframe.removeAttribute('data-km-custom-widget-position');
+    }
+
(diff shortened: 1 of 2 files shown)


webplugin/js/app/mck-app.js
```diff
@@ -399,6 +399,58 @@ function ApplozicSidebox() {
...
+    function resetWidgetCustomPosition(kommunicateIframe) {
+        widgetPositionProperties.forEach(function (property) {
+            kommunicateIframe.style.removeProperty(property);
+        });
+        kommunicateIframe.removeAttribute('data-km-custom-widget-position');
+    }
+
+    function getWidgetCustomPositionProperties(widgetSettings) {
+        var horizontalPositionProperty =
+            widgetSettings.position === KommunicateConstants.POSITION.LEFT ? 'left' : 'right';
```
devashishmamgain: why this and not below one?

 var horizontalPositionProperty = widgetSettings.position

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -804,11 +804,7 @@ const firstVisibleMsg = {
             } else {
                 kommunicateIframe.classList.add('km-iframe-dimension-no-popup');
             }
-            if (!PRE_CHAT_LEAD_COLLECTION_POPUP_ON) {
-                kommunicateCommons.adjustIframeHeightForLayout(kommunicateIframe);
-            } else {
-                kommunicateIframe.style.height = '';
-            }
+            kommunicateCommons.adjustIframeHeightForLayout(kommunicateIframe);
```
devashishmamgain, line 807: what was the purpose of previous code if (!PRE_CHAT_LEAD_COLLECTION_POPUP_ON) ?
and why is that check not required now?

ayush-kommunicate: Sir, previously the pre-chat lead collection popup was using a fixed  height. When the iframe height changed dynamically, the pre-chat form overflow outside the iframe.
This was like a bug in the existing implementation. Now, the pre-chat lead collection popup also follows the same dynamic height adjustment logic, so the iframe height is updated correctly and the overflow issue is resolved. And i also checked it is not causing any issue.

webplugin/js/app/mck-app.js
ayush-kommunicate: Changed

devashishmamgain (approved):

PR #1535, opened 2026-06-09
TDH-3621-Duplicate-user-with-same-email

ayush-kommunicate:
- In the Pre-Lead Chat flow, when a user submits a form, we currently normalize the email but not the user_name, even when the user_name is the same as the email.
- For example, if a user enters Ayush@gmail.com, it is saved as Ayush@gmail.com in user_name and ayush@gmail.com in email. Later, if a user enters AYush@gmail.com, it is saved as a different user_name but the same normalized email (ayush@gmail.com).
- Since we check user_name for uniqueness, both values are treated as different, resulting in multiple users being created with the same email.

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pre-chat login email handling by consistently trimming whitespace and converting emails to lowercase.
  * The normalized email is now used for session persistence and for the email value provided during pre-chat initialization, ensuring users are correctly recognized even when emails are entered with extra spaces or different capitalization.

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -5026,13 +5026,13 @@ const firstVisibleMsg = {
                     }
                     var normalizedEmail = email ? $applozic.trim(email).toLowerCase() : null;
                     if (normalizedEmail) {
-                        const userIdForCookie = anonymousUserIdForPreChatLead
+                        const userIdForLocalStorage = anonymousUserIdForPreChatLead
                             ? userId
                             : normalizedEmail;
-                        userId = userIdForCookie;
+                        userId = userIdForLocalStorage;
                         kmLocalStorage.setLocalStorage({
                             name: KommunicateConstants.COOKIES.KOMMUNICATE_LOGGED_IN_ID,
-                            value: userIdForCookie,
+                            value: userIdForLocalStorage,
                             expiresInDays: 30,
                         });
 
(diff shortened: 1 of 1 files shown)


devashishmamgain (changes requested):
@ayush-kommunicate @aashish-bhatt i think this should get handled from backend side, similar issue might happen for mobile sdks or direct apis calls

AashishBhatt5:
Yes, @Ayush Khanduri ***@***.***> in chat server before saving it
to db change email to lowercsse

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -5024,11 +5024,12 @@ const firstVisibleMsg = {
                             userId = contactNumber;
                         }
                     }
-                    if (email) {
-                        const userIdForCookie = anonymousUserIdForPreChatLead ? userId : email;
-
+                    var normalizedEmail = email ? $applozic.trim(email).toLowerCase() : null;
+                    if (normalizedEmail) {
+                        const userIdForCookie = anonymousUserIdForPreChatLead
```
devashishmamgain: are we still using cookies? if not, please rename it to something else to avoid confusion

ayush-kommunicate: Now we are storing this in Local storage. Changed variable name.

devashishmamgain (approved):

PR #1538, opened 2026-06-17
TDH-3885:Fix broken autosuggestions payload implementation

adityaKommunicate:
-Fix the Web SDK autosuggestion dropdown so it appears correctly.

### PR Checklist

### How was the code tested?

-

-

-The autosuggestion dropdown menu was reused across prompts and was being hidden with the n-vis class after the first interaction. On the next autosuggestion prompt, the typeahead show() flow called .show() but did not remove n-vis, so the menu stayed hidden even though suggestions were rendered and the component considered itself open.

### Screenshot
- 

## Bug Fixes
* Fixed an issue where search suggestions dropdown menu would not display properly when the autosuggest feature was re-initialized on input fields with existing typeahead instances.

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -9478,6 +9478,10 @@ const firstVisibleMsg = {
             });
 
             _this.populateAutoSuggest = function (params) {
+                var existingTypeahead = $mck_autosuggest_search_input.data('mcktypeahead');
+                if (existingTypeahead && existingTypeahead.$menu) {
+                    existingTypeahead.$menu.removeClass('n-vis');
+                }
                 $mck_autosuggest_search_input.mcktypeahead({
                     order: 'desc',
                     hint: false,
--- a/webplugin/lib/js/mck-ui-widget.min.js
+++ b/webplugin/lib/js/mck-ui-widget.min.js
@@ -672,16 +672,16 @@
                 return this.$menu.css({
                     bottom: 50 + 'px',
                     left: 0 + 'px'
-                }),this.$menu.show(), this.shown = !0, this
+                }).removeClass("n-vis").show(), this.shown = !0, this
             }else{
                 return this.$menu.css({
                     top: t.top + t.height,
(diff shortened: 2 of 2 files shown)


webplugin/lib/js/mck-ui-widget.min.js
```diff
@@ -672,16 +672,16 @@
                 return this.$menu.css({
                     bottom: 50 + 'px',
                     left: 0 + 'px'
-                }),this.$menu.show(), this.shown = !0, this
+                }).removeClass("n-vis").show(), this.shown = !0, this
```
devashishmamgain: .min.js files are library files, they are not supposed to tbe touched. whatever changes are required, need to be done in .js files

devashishmamgain (changes requested):

webplugin/lib/js/mck-ui-widget.min.js
adityaKommunicate: okay

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -9478,6 +9478,10 @@ const firstVisibleMsg = {
             });
 
             _this.populateAutoSuggest = function (params) {
+                var existingTypeahead = $mck_autosuggest_search_input.data('mcktypeahead');
+                if (existingTypeahead && existingTypeahead.$menu) {
+                    existingTypeahead.$menu.removeClass('n-vis');
```
devashishmamgain, line 9483: will this be remove n-vis or .hide() as the library is using show hide and not vis, n-vis?

is this change tested?

adityaKommunicate: the autosuggest failure was caused by the mcktypeahead reused menu retaining the n-vis class from the previous cycle even though the new autosuggest payload and menu items were present.fix clears n-vis. I tested the changes locally its working
image
image

devashishmamgain (approved):

thakaresakshi121 (approved):

PR #1543, opened 2026-06-29
Stream empty fix

thakaresakshi121:
-Root cause: addTokenizeMsg returned permanently when msg.index was not exactly the next expected index, so later chunks never rendered. Now the service buffers chunks by index, renders contiguous chunks in order, and flushes remaining buffered chunks on stream completion so out-of-order delivery still displays the full received message.

### PR Checklist

### How was the code tested?

-

-

-

### Screenshot
- 

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -7891,8 +7891,7 @@ const firstVisibleMsg = {
                 // GEN AI BOT
 
                 if (msg.tokenMessage && !msgThroughListAPI) {
-                    // message not from the sockets
-                    document.getElementById('mck-text-box').setAttribute('contenteditable', true);
+                    genAiService.enableTextArea(false);
                 }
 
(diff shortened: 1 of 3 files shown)


webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -7901,8 +7891,7 @@ const firstVisibleMsg = {
                 // GEN AI BOT
 
                 if (msg.tokenMessage && !msgThroughListAPI) {
-                    // message not from the sockets
-                    document.getElementById('mck-text-box').setAttribute('contenteditable', true);
+                    genAiService.enableTextArea(false);
```
ayush-kommunicate, line 7894: Is this not block the text area again?

ayush-kommunicate (changes requested):

webplugin/js/app/mck-sidebox-1.0.js
thakaresakshi121: it will block it again for next streaming message

ayush-kommunicate (approved):

PR #1549, opened 2026-07-09
km-input-voice-metadata-issue

adityaKommunicate:
- Fix voice-mode message metadata so CTA/button-triggered sends also include `KM_INPUT_TYPE: "VOICE"` when voice chat is active.

### PR Checklist

### How was the code tested?
- 

-

- 

### Screenshot
- N/A for this change. The expected verification is in the request payload: button/CTA sends in active voice mode should now include both `KM_CHAT_CONTEXT` and `KM_INPUT_TYPE: "VOICE"`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of voice input so outgoing chat messages correctly reflect voice mode when active.
  * Ensures messages are not unintentionally tagged: if a message has no text or includes attachments, existing metadata is preserved and no voice tagging is applied.

### DIFF
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -5751,6 +5751,7 @@ const firstVisibleMsg = {
                     resolvedMetadata.KM_INPUT_TYPE ||
                     typeof mckVoice === 'undefined' ||
                     !mckVoice ||
+                    typeof mckVoice.isVoiceModeActive !== 'function' ||
                     !mckVoice.isVoiceModeActive() ||
                     !messagePxy.message ||
                     messagePxy.fileMeta
--- a/webplugin/js/app/mck-sidebox-1.0.js
+++ b/webplugin/js/app/mck-sidebox-1.0.js
@@ -5745,6 +5745,22 @@ const firstVisibleMsg = {
                 );
             };
 
+            _this.updateMessageMetadataForVoiceInput = function (messagePxy, metadata) {
+                var resolvedMetadata = metadata || {};
+                if (
+                    resolvedMetadata.KM_INPUT_TYPE ||
+                    typeof mckVoice === 'undefined' ||
+                    !mckVoice ||
+                    !mckVoice.isVoiceModeActive() ||
+                    !messagePxy.message ||
+                    messagePxy.fileMeta
+                ) {
+                    return resolvedMetadata;
+                }
+                resolvedMetadata.KM_INPUT_TYPE = 'VOICE';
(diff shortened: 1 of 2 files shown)


thakaresakshi121 (dismissed):

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -5745,6 +5745,23 @@ const firstVisibleMsg = {
                 );
             };
 
+            _this.updateMessageMetadataForVoiceInput = function (messagePxy, metadata) {
+                var resolvedMetadata = metadata || {};
+                if (
+                    resolvedMetadata.KM_INPUT_TYPE ||
+                    typeof mckVoice === 'undefined' ||
+                    !mckVoice ||
+                    typeof mckVoice.isVoiceModeActive !== 'function' ||
```
devashishmamgain: why is function check required?

devashishmamgain (changes requested):

webplugin/js/app/mck-sidebox-1.0.js
adityaKommunicate: Not required removed

devashishmamgain (approved):

webplugin/js/app/mck-sidebox-1.0.js
```diff
@@ -5745,6 +5745,22 @@ const firstVisibleMsg = {
...
+                    resolvedMetadata.KM_INPUT_TYPE ||
+                    typeof mckVoice === 'undefined' ||
+                    !mckVoice ||
+                    !mckVoice.isVoiceModeActive() ||
+                    !messagePxy.message ||
+                    messagePxy.fileMeta
+                ) {
+                    return resolvedMetadata;
+                }
+                resolvedMetadata.KM_INPUT_TYPE = 'VOICE';
```
devashishmamgain: these checks looks wrong, if voice mode is active, only then voice should get set. here if mckVoice is not present, lets say voice is disabled or voice file is not included then also it will set KM_INPUT_TYPE = 'VOICE'

devashishmamgain (changes requested):

webplugin/js/app/mck-sidebox-1.0.js
adityaKommunicate: checking

adityaKommunicate: now set KM_INPUT_TYPE only when mckVoice exists and isVoiceModeActive() returns true.

devashishmamgain (approved):

PR #1560, opened 2026-09-09
[TDH-3904] Support localized trigger buttons

kandpal025:
## Jira
- [TDH-3904](https://kommunicate.atlassian.net/browse/TDH-3904)

## Summary
- Extend startup trigger buttons (`kommunicateSettings.quickReplies`) to support rich quick-reply-style objects in addition to the existing string array.
- Render localized button text from `title`/`name`, send the configured `message`, preserve optional `replyMetadata`, and apply `updateLanguage`/`languageCode` when the visitor clicks the trigger button.
- Keep existing string trigger buttons on the same text-box submit path for backward compatibility.

## Root Cause
Trigger buttons are configured through the same `quickReplies` setting documented for the web widget, but the startup renderer treated every item as a plain string and sent `innerText` through the text box. That meant trigger buttons could not use the localized display text, reply metadata, or language update fields already supported by rich-message quick replies.

## Validation
- `git diff --check`
- Static diff check: PR contains only `webplugin/js/app/kommunicate-ui.js`

[TDH-3904]: https://kommunicate.atlassian.net/browse/TDH-3904

### DIFF
--- a/webplugin/js/app/kommunicate-ui.js
+++ b/webplugin/js/app/kommunicate-ui.js
@@ -1553,17 +1553,59 @@ KommunicateUI = {
         }
     },
     loadQuickReplies: function (quickReplies) {
+        var getQuickReplyValue = function (quickReply, keys) {
+            if (!quickReply || typeof quickReply !== 'object') {
+                return '';
+            }
+            for (var keyIndex = 0; keyIndex < keys.length; keyIndex++) {
+                if (quickReply[keys[keyIndex]]) {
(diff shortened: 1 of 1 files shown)


kandpal025:
Closing this PR because the branch name contains `/`, and the Cloud Build Sentry release step uses the branch name as the release version. Sentry rejects slashes in `<VERSION>`, so this branch fails before the build completes.

Replacement PR with the same commit on a slash-free branch: #1561

